PR #15, opened 2015-02-05
Delete unneccessary files

carsongee:
This adds in the standard github python `.gitignore` file (and adds to the existing one).  I also deleted the zip-js and slideshowjs folders.  zip.js wasn't included anywhere and slideshowjs still has the runtime of library in the html_app/js folder, so it should be okay to delete the downloaded full project  I deleted the scb_rest folder as it was a completely empty django app, and deleted the build folder as it dynamically buillt  @annagav @skini @ferdi


Ferdi:
Great !  I dont have write access to this repo but it looks merge-able to me

PR #18, opened 2015-02-12
Fix outdated fixtures to match the models

bdero:


carsongee:
Lovely, are we finally out of 500's with this?

bdero:
Yes! :octopus:

PR #14, opened 2015-01-07
New docs folder in starcellbio_html develop branch

cshubert:
This includes lots of StarCellBio documentation as well as a new Developer's Guide.

PR #27, opened 2015-02-19
Port forward 8100 with vagrant.

bdero:
This is useful when private_networks don't work properly. (When using the Cisco VPN client, for instance)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -25,7 +25,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # Create a private network, which allows host-only access to the machine
   # using a specific IP.
   config.vm.network "private_network", ip: "192.168.33.200"
-  config.vm.network :forwarded_port, guest: 8000, host: 8000
+  config.vm.network :forwarded_port, guest: 8100, host: 8100
   # Create a public network, which generally matched to bridged network.
   # Bridged networks make the machine appear as another physical device on
   # your network.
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -25,7 +25,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # Create a private network, which allows host-only access to the machine
   # using a specific IP.
   config.vm.network "private_network", ip: "192.168.33.200"
-
+  config.vm.network :forwarded_port, guest: 8000, host: 8000
```
carsongee: maybe pick something more unique, this will compete with devstack at least, and probably others

carsongee:
After port change, and something in the README or documented somewhere (or the developer doc on @cshubert's PR), not sure what makes the most sense.

bdero:
What do you think of something like 6060? There's currently no readme in develop (that I know about, at least). It also looks like there isn't one in Chuck's repo. One of my hanging branches is for adding a README.md, but I ditched it because I didn't want to describe StarCellBio wrong.

I think I'll give it a shot and do a PR to see what you guys think. It would be good to have vagrant provisioning as well as info about the watcher script included in the readme.

carsongee:
I was referring to this: https://github.com/cshubert/starcellbio_html/blob/develop/docs/DeveloperGuide.rst
We may want to pull that out of that big PR https://github.com/starteam/starcellbio_html/pull/14 just to get that piece in

carsongee:
Also 6060 is x11, but 8100 is empty: http://www.speedguide.net/ports.php?filter=8100 or 8200, or probably a lot of others

bdero:
Alright, #30 is a PR for the developer guide document. I couldn't do an amended cherry-pick since the blame was made up of several commits.

bdero:
8100 sounds cool with me. Using that instead. Thanks for checking up on that port and making suggestions; I used a different (less thorough) web resource to look up port usage.

All comments addressed.

PR #28, opened 2015-02-20
Remove and ignore all generated .soy.js and .gss.css files

bdero:
I verified that they're all regenerated. There were actually a couple of versioned generated filed here that were no longer used and no longer have their respective source files versioned.


.gitignore
```diff
@@ -66,4 +66,5 @@ html_app/index.html
 .vagrant/
 static/
 
-
+*\.soy\.js
+*\.gss\.css
```
carsongee: Can we not just ignore folders called `gen`? or are there things we want to keep in them?

bdero:
Yep, just tried it and committed; same exact result!

carsongee:
cool, :tangerine:

bdero:
:clap:

PR #29, opened 2015-02-20
Settings configurable with untracked config and environment variables

bdero:
- The untracked YAML config `settings.yml` merges and overrides defaults if it exists.
- Environment vars that start with `SCB_` get the last say. I.e. `SCB_DB_NAME` turns into `DB_NAME`.

### DIFF
--- a/StarCellBio/settings.py
+++ b/StarCellBio/settings.py
@@ -62,7 +62,7 @@ USE_TZ = True
 
 # Absolute filesystem path to the directory that will hold user-uploaded files.
 # Example: "/home/media/media.lawrence.com/media/"
-#MEDIA_ROOT = '/tmp/static/'
+# MEDIA_ROOT = '/tmp/static/'
 
 # URL that handles the media served from MEDIA_ROOT. Make sure to use a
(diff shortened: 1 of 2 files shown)


StarCellBio/settings.py
```diff
@@ -195,4 +193,46 @@
...
+    with open(YAML_CONFIG) as f:
+        y = yaml.load(f)
+        if y is not None:
+            globals().update(y)
+
+# Override settings with SCB_ environment variables
+scb_env_overrides = {
+    key[4:]: value
+    for key, value in os.environ.iteritems()
+    if key.startswith("SCB_")
```
carsongee: This is a bit terse for me as a dictionary comprehension, and it arbitrarily makes it python 2.7 only (and the servers running this are 2.6).

carsongee:
After comment and looks like a rebase, :kimono:

bdero:
Good point, I'll fix that up.

bdero:
Alrighty, I separated the list comprehension out to a map and filter. Merging this.

PR #31, opened 2015-02-25
Use watchdog for static file watching

bdero:
Uses docopt for argument parsing and watchdog for observing static files.
Lots of cleanup was done here, but I made a point to not rewrite any logic in `processor()` for this PR, so there are still a bunch of pylint violations.

After this PR, the next steps are to have it run automatically in the Vagrant box, as well as to clean up `processor()` and the `index.html` generation.


bdero:
The file rename is causing the combined diff to look silly. Would recommend looking at the individual commits; if it's a hassle, I can cut away the rename commit temporarily for review.

requirements.txt
```diff
@@ -23,3 +23,5 @@ selenium==2.35.0
 six==1.3.0
 urwid==1.1.1
 wsgiref==0.1.2
+watchdog==0.8.3
+docopt==0.6.2
```
carsongee: I'd prefer >= until we know something breaks

html_app/build.py
```diff
@@ -0,0 +1,257 @@
+#!/usr/bin/env python
+
+"""The static asset pipeline for StarCellBio.
```
carsongee, line 3: Hadn't seen docops before, not sure it saves you much typing time over argparse, but it is a pretty neat concept. How do you specify integer/boolean/file args, or do nargs, or choices? It also is maybe a little too magical for me, but is fine to use here.

carsongee:
Can you squash everything but the file name change so the diff is readable?  It is really hard to tell what you changed vs what was there already, but if everything but the rename is lumped in a commit I can read the commit that has all the changes to the original.

html_app/build.py
```diff
@@ -0,0 +1,257 @@
...
+        level=logging.INFO,
+        format='[%(asctime)s] %(levelname)s: %(message)s',
+        datefmt='%Y-%m-%d %H:%M:%S'
+    )
+
+    if args['watch']:
+        watch(".")
+    else:
+        # Run a build
+        if soy_only or gss_only:
```
carsongee: I'd generally prefer not to add options that aren't implemented.  Are you planning to implement these soon?

bdero: Ah, my bad, I intended to remove that

html_app/build.py
bdero, line 3: Accepted arguments are always strings, and I don't think there's a way to validate that the argument passed was an integer. The example on http://docopt.org/ shows choices.

Here's the docstring from that silly deckgen script as an example including nargs and type implication:

``` manual
Standalone slide deck collection builder.

Usage:
  deckgen.py build [-s SRC] [-d DEST] [-t TEMP] [NAME...]
  deckgen.py watch [-s SRC] [-d DEST] [-t TEMP]
  deckgen.py -h | --help

Arguments:
  name   The name of a slide deck to generate. Accompanying markdown should be
         located in the source directory. For example, if the markdown file is
         named "slides.md", the name "slides" should be passed.
         If no names are passed, the default behavior is to run against ALL
         markdown files in the source directory

Options:
  -h --help               Show this help screen.
  -s --source=SRC         Source markdown directory [default: markdown].
  -d --destination=DEST   Destination HTML directory [default: decks].
  -t --template=TEMP      HTML template to use [default: template.html].
```

That being said, docopt is indeed waaaay too magical and doesn't scale for really complicated CLIs and situations requiring a lot of fine grained control, but I figured it would be okay here.

requirements.txt
bdero: Ah, okay, will do.

bdero:
Alrighty, I addressed the comments and squashed + got rid of the rename commit. Ready for review.

html_app/watch.py
```diff
@@ -1,125 +1,248 @@
...
-    js_join = js_prefix + "starcellbio.app.js" + js_suffix + js_join
-    js_join = js_prefix + "js/soyutils.js" + js_suffix + js_join
-    js_join = js_prefix + "js/tinymce.min.js" + js_suffix + js_join
-    js_join = js_prefix + "js/jquery.tinymce.min.js" + js_suffix + js_join
-    js_join = js_prefix + "js/jquery-1.7.2.min.js" + js_suffix + js_join
-    js_join = js_join + js_prefix + "master_model.js" + js_suffix
-
-
-    return html_prefix + css_join + js_join + html_suffix
+    global css, js
```
carsongee: I know you aren't changing stuff here, so it won't affect the review, but what the what

bdero: same

html_app/watch.py
```diff
@@ -1,125 +1,248 @@
...
+"""The static asset pipeline for StarCellBio.
+
+Usage:
+  build.py
+  build.py watch
+  build.py -h | --help
+
+Options:
+  -h --help       Show this help screen.
+  -s --soy-only   Translate all "soy" files to JavaScript.
```
carsongee: remove these too, since you pulled the code to handle them

carsongee:
Other than that last comment :cow2:

carsongee:
oh, and rename it again

html_app/watch.py
bdero: oops!

bdero:
Alright, ready to go. :golf:

PR #36, opened 2015-02-27
Release v0.6.0?

bdero:
This is going off of the latest semantic-looking version here: https://github.com/starteam/starcellbio_html/releases/tag/v0.5

There are several release tags that are dates, like `V20140925`, for instance. Some releases follow another pattern as well.
We should probably discuss and think about versioning, and perhaps a roadmap for StarCellBio moving forward.

This probably shouldn't be merged until we have a clean production environment ready to go.


carsongee:
Totally un-reviewable, plus merge conflicts, but hey I'm :+1:  for 0.6.0 :hammer: . Can you maybe point to where that version is specified in the code?

bdero:
Using release branch instead.. Ditching master.

Issue #37, opened 2015-02-27
SUTD #1 Add Samples button doesn't work


pdpinch:
> I started reviewing SUTD exercise #1 as this is a top priority for this spring. Unfortunately, I can't move past the Experiment setup page as the 'Add Samples' button isn't working. I then logged into the course on which this exercise is based (course code 7.06_Spring_2014) and was able to confirm that I can still perform experiments in this original exercise. Can you look into SUTD exercise #1 for us? If you need more information on this exercise and/or the work that was done to get it ready for SUTD, either Lourdes or I can fill you in.

pdpinch:
@bdero -- when you can -- can we get a deployment of develop (4a6697a876777e7239a0d84dd6dfa4eb47e3f046) so we can test this?

bdero:
This was deployed this morning. Is this issue all set now?

pdpinch:
I'm just waiting for confirmation from Alison or Lourdes.

PR #45, opened 2015-03-03
fixed assignment setup and add_sample_dialog

annagav:


html_app/master_model.js
```diff
@@ -15881,15 +15881,13 @@ var fix_assignment_models =  function (assignment_result) {
...
-				['Background Information',scb_assignment_sutd_intro_biology.page2()],
-				['Questions', scb_assignment_sutd_intro_biology.page3()]
+				['Getting Started', 'Please refer to your instructor for detailed instructions.']
 				];
 			sutd_intro_biology.template.experiment_setup = scb_assignment_sutd_intro_biology.experiment_setup();
 
 			sutd_intro_biology.template.ui.experiment_setup.actions.push({
 				name: 'ADD SAMPLES',
-				open: 'mit706s14ps2.setup',
+				open: 'mit706s14.setup',
```
pdpinch, line 15890: So, are these names completely arbitrary, or do different assignments have dependencies on the same data object, mit706s14?

pdpinch:
Tested on my vagrant and it works, as long as you aren't using an account with old data -- kind of scary, but something to address at another time.

pdpinch:
fixes #37

html_app/master_model.js
annagav, line 15890: Usually different assignments do not have dependencies on the same data object (for setup and instructions). But in this case, since we are recycling previous assignment it is easier to use existing object.

PR #63, opened 2015-03-06
Fixes instruction page for the assignment

annagav:
- Disabled arrow buttons if there is only one instruction.
- Changed the instruction text of the sutd2 assignment.

### DIFF
--- a/html_app/master_model.js
+++ b/html_app/master_model.js
@@ -16942,9 +16942,6 @@ var fix_assignment_models =  function (assignment_result) {
 		}
 		if(assignment_result[x].id == 'sutd_intro_biology'){
 			sutd_intro_biology = assignment_result[x];
-//			sutd_intro_biology.template.instructions = [
-//				['Getting Started', 'Please refer to your instructor for detailed instructions.']
-//				];
 			sutd_intro_biology.template.experiment_setup = scb_assignment_sutd_intro_biology.experiment_setup();
 
 			sutd_intro_biology.template.ui.experiment_setup.actions.push({
--- a/html_app/master_model.js
+++ b/html_app/master_model.js
@@ -9612,7 +9612,7 @@ var __sutd_intro_biology = {
     experiments: {},
         template: {
             instructions: [
-            	['Goal & Introduction','Here come instructions when we build them']
+           	    ['Goal & Introduction','Please refer to your instructor for detailed instructions.']
             	],
             ui: {
                 experimental_design: {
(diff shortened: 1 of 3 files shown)


pdpinch:
It's Ok that these are combined in one PR, since they are part of the same view. But I would have preferred to have them separate as your 1st bullet is a template change for all assignments and your 2nd bullet is a content change that only applies to sutd2.

pdpinch:
Whoa, why so many changes to AssignmentsView.js ?

annagav:
I did code reformat to fix some of the alignment I ruined.

html_app/master_model.js
```diff
@@ -16942,9 +16942,9 @@ var fix_assignment_models =  function (assignment_result) {
 		}
 		if(assignment_result[x].id == 'sutd_intro_biology'){
 			sutd_intro_biology = assignment_result[x];
-			sutd_intro_biology.template.instructions = [
-				['Getting Started', 'Please refer to your instructor for detailed instructions.']
-				];
+//			sutd_intro_biology.template.instructions = [
+//				['Getting Started', 'Please refer to your instructor for detailed instructions.']
+//				];
```
pdpinch: Can we just delete these lines?

html_app/ui/assignments.soy
```diff
@@ -78,19 +78,21 @@ Assignment description
...
 				{$section[1] |noAutoescape}</div>   
-        	{/foreach} 
+        	{/foreach}
+        	{if length($assignments.selected.template.instructions)>1}
         	<div class='scb_s_assignments_bottom_scroll'>
    				<div class = 'scb_s_assignment_header_img_left scb_assignments_bottom_arrow_left scb_s_assignment_header_img_left_inactive'  assignment_id='{$assignments.selected.id}'   role='button' aria-label='Previous Section'></div>
    				<div class='scb_s_assignments_button_divider'></div>
 				<div class = 'scb_s_assignment_header_img_right scb_assignments_bottom_arrow_right'  assignment_id='{$assignments.selected.id}'  role='button' aria-label='Next Section'></div>
 			</div>
+			{/if}
```
pdpinch, line 95: I think this would be easier to read if these two lines (94-95) were de-indented to match lines 89-90.

pdpinch:
:+1: once the comments are addressed.

html_app/master_model.js
annagav: That's right. Ill do it now.

PR #79, opened 2015-03-09
sutd2: added more images to slide_parser

annagav:


pdpinch:
What's this for?

It seems like there should be way to do this that involves less typing.

laleman:
I do not know what this is referring to.

pdpinch:
Sorry, that was a question for @annagav 

> What's this for?
> It seems like there should be way to do this that involves less typing.

annagav:
This slide_parser tells the microscopy parser to assign these images to a specific treatment.

PR #81, opened 2015-03-09
FACS: choose condition. Fixes #72

### LINKED ISSUE #72: SUTD Exercise #2 - Flow Cytometry: "Conditions" should be left blank for this particular exercise
An example of how it should look like can be found in 2014 7.06 Assignment #2
screen shot 2015-03-08 at 9 28 59 pm

annagav:
If "facs kind" is 'whole' then condition is blank.
For other "facs kinds":  
- if there is only one condition then a checked radio button for that condition is displayed,
- for more conditions a dropdown menu is displayed.


pdpinch:
How hard would it be to omit the conditions column altogether? (as suggested in #72)

pdpinch:
Ok. We'll merge this for now with the empty column & revisit later once we resolve the soy version issue.

PR #82, opened 2015-03-09
sutd2: fixed close add sample dialog button. Fixes #68

### LINKED ISSUE #68: SUTD Exercise #2 - The "X" that closes the window in the "Add Samples" pop-up window does not work

annagav:
fixes #68


pdpinch:
This works, but why do classes for basic ui elements need to be unique across templates? Couldn't this be "inner_dialog_title_close" and then require less customization?

PR #85, opened 2015-03-09
Switch contact form to use form validation and django e-mail

### LINKED ISSUE #84: Email uses subprocess `/usr/sbin/sendmail`
This will likely cause e-mailing not to work in production and should be replaced with the django e-mail facilities.

carsongee:
Still defaults to star@mit.edu, but is configurable via django settings
Closes #84


StarCellBio/views.py
```diff
@@ -1,23 +1,22 @@
...
-import json
-import pudb
+from django.template.loader import render_to_string
+
 from backend.models import Assignment, StudentAssignment, Course, UserCourse
-import StringIO
-from django.core.files.base import ContentFile
+from StarCellBio.forms import ContactForm
 import StarCellBio.settings
-import os
```
pdpinch, line 16: were there really that many unnecessary imports?

carsongee, line 16: pyFlakes suggested that is so

bdero:
Tested working and all looks well :musical_score:

Issue #62, opened 2015-03-06
The order of included scripts when running build.py is not deterministic


bdero:
This causes the need for manual reordering during deployment since the order of the included scripts is super important.

https://github.com/starteam/starcellbio_html/blob/639381461d4017a0648acc8055e714a36f93123c/html_app/build.py#L94-L98

carsongee:
Any objections to just making all the JS dependencies a flat hard coded list, or do we want to keep the dynamic asset discovery and just figure out what dep is breaking things?

pdpinch:
I haven't used the dynamic asset discovery yet, so I'm fine with a hard-coded list. But we should also check with @annagav

carsongee:
It is currently ~112 files, so a long list, but I don't know how often it is changed

annagav:
No objections

PR #91, opened 2015-03-10
Optimize asset pipeline for CSS and Javascript

carsongee:
Create deterministic jasvascript loading
Minify all JS into single file with `build.py -p`
Minify all CSS into single file with `build.py -p`


carsongee:
Closes https://github.com/starteam/starcellbio_html/issues/62 and should help with performance in production (about ~1.5 seconds faster when loaded locally vs. with individual files)

Vagrantfile
```diff
@@ -26,7 +26,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # using a specific IP.
   config.vm.network "private_network", ip: "192.168.33.200"
   config.vm.network :forwarded_port, guest: 8100, host: 8100
-
+  config.vm.provider "virtualbox" do |vb|
+    vb.customize ["modifyvm", :id, "--memory", "1024"]
+  end
```
pdpinch, line 31: Is this increasing the memory on the VM?

pdpinch:
@annagav @bdero please review

Vagrantfile
carsongee, line 31: Yeah, It probably isn't strictly required, but I OOMed the box doing the minify with uglify-js

bdero:
Thanks a lot for this. Tried it out and works in the vm out of the box. Waaaaaay faster with the -p flag. :money_with_wings:

PR #104, opened 2015-03-14
facs: fixed gate labeling. fixes #74

### LINKED ISSUE #74: Flow Cytometry: Bisector gate "Apply All" button does not apply labels to all graphs
When you apply the bisector gate to all of the samples only the original sample used to create the gates has the "a" and "b" labels on top of the gates

In the example provided, the bisector gate was used for the "Control siRNA" sample. After the button "Apply All" has been checked, all of the other samples should have the same gates applied to them with the same labels. I am providing an example of siRNA #1 sample so that you can see that the gates are missing the "a" and "b" labels from gates. 
screen shot 2015-03-08 at 9 41 22 pm
screen shot 2015-03-08 at 9 41 30 pm

annagav:


pdpinch:
I discovered an unrelated issue, #111. Otherwise this looks good to me. :+1:

PR #113, opened 2015-03-17
facs: progress bar fixed, video added

### LINKED ISSUE #105: Add an "In the Lab" feature to the flow cytometry pages
The flow cytometry pages currently do not have an in the lab feature box, but we have two animations to be featured in an 'in the lab' box. The In the lab box should definitely appear on the Sample prep page, and also on the run/analyze pages if there's room (like on the microscopy pages). Once this is done, then the flow cytometry animation (http://youtu.be/2P7YsJ0Zkio) and the fluorescence animation (http://youtu.be/pFtO0xxJL9s) can be added to the box.

annagav:
FACS: Added IN THE LAB videos. Had to move the progress bar to the left to fit the video as it is done in microscopy. Fixed previous issue of the progress bar misalignment. Fixes #105.


pdpinch:
looks good

PR #119, opened 2015-03-19
facs: switching to pseudo logarithmic scale.

annagav:


html_app/components/FACSModelFactory.js
```diff
@@ -64,7 +63,7 @@ scb.components.FACSModelFactory = function scb_components_FACSModelFactory(model
              */
         }
         self.shape_to_data = function (state) {
-            var shape = state.shape;
+            var shape = ('' + state.shape).toLowerCase();
```
pdpinch, line 66: What's the `('' + state.shape)` for? Does that coerce state.shape into a string?

html_app/components/FACSModelFactory.js
```diff
@@ -319,7 +313,17 @@ scb.components.FACSModelFactory = function scb_components_FACSModelFactory(model
                         }
                     },
 
-//                    tickFormatter: function (v, axis) {return "10^" + (Math.round( Math.log(v)/Math.LN10)).toString();}, //(Math.round( Math.log(v)/Math.LN10)).toString().sup();},
+                    tickFormatter: function (v, axis) {
+                        if (template.model.facs.scale && template.model.facs.scale.indexOf('pseudo') > -1) {
+                                return "10^"+Math.round(v/template.model.facs.ticks[0]);
+                        } else if (template.model.facs.scale && template.model.facs.scale.indexOf('log') > -1) {
```
pdpinch, line 319: I'm curious: what are the possible values of template.model.facs.scale ? Are those documented anywhere?

annagav, line 319: For Exercise2 and 3 the value of scale is  'pseudo-log' (it can be any string containing 'pseudo'), the rest of new exercises have 'linear' for linear scale. Originally, it was 'log' for logarithmic, but since we are probably not going to use that, the second if can be omitted.

html_app/components/FACSModelFactory.js
annagav, line 66: That's right. I moved this transformation from the if statement to the declaration.

pdpinch, line 66: I think `String(state.shape).toLowerCase()` or `state.shape.toString().toLowerCase()` would be clearer -- assuming they work.

annagav, line 66: This was not my piece of code. I just moved it from one place to another.

pdpinch:
This looks good, but we need to open an issue for the gate (x) values -- % Cells seems fine, but the treatment range uses the linear values instead of the logarithmic values.

PR #123, opened 2015-03-20
NSF Exercise 2 FACS complete.

annagav:


pdpinch:
It looks like you'll need to change the Analysis tools (single gate and bisector gate) to work with this new scale. When I try to create a bisector gate, it appears to be reading the linear location of my click, but then it places it on the graph using the logarithmic scale. 

Dragging

after mouseup

annagav:
The scale does not look right. It is linear instead of logarithmic.

pdpinch:
Ok, now that I'm looking at the right branch, this looks good.

PR #129, opened 2015-03-23
changes for django upgrade to 1.7

annagav:
- changes for django upgrade to 1.7
- backend initial migration
- changed django version in requirements.txt


pdpinch:
It seems like we need some changes to https://github.com/starteam/starcellbio_html/blob/develop/requirements.txt ?

StarCellBio/settings.py
```diff
@@ -128,16 +128,9 @@
...
-    'django.contrib.auth',
-    'django.contrib.contenttypes',
-    'django.contrib.sessions',
-    'django.contrib.sites',
-    'django.contrib.messages',
-    'django.contrib.staticfiles',
-    # Uncomment the next line to enable the admin:
-    'django.contrib.admin',
     # Uncomment the next line to enable admin documentation:
     'django.contrib.admindocs',
```
pdpinch, line 135: This doesn't seem to be working, due to a dependency on docutils. Do we still want it?

carsongee, line 135: I've never really used it

bdero:
Hmm, I just did a `vagrant destory; vagrant up` and ran into a classic django upgrade problem:
screenshot

Has this been tested on a fresh vagrant box?

requirements.txt
```diff
@@ -1,4 +1,4 @@
-Django==1.4.18
+Django==1.7.7
```
carsongee, line 1: Did you also try upgrading all the other django dependencies?  Chances are that they needed to be updated to support django 1.7 as well

carsongee, line 1:
```
django-allauth==0.19.1
django-bootstrap-form==3.2
django-filter==0.9.2
django-tastypie==0.12.1
django-uni-form==0.9.0
djangorestframework==3.1.1
```

are all the latest for the django-requirements

annagav, line 1: I did try upgrading django-allauth but it seemed to break more things. What would be the strategy for upgrading other django dependencies?

carsongee, line 1: since we don't have tests, there isn't much to do besides upgrade and try everything out.  Then fix any 500s that come up.

carsongee:
So, after a little fiddling, I'm not having this error.  Using the updated requirements here, removing every line in `auth/views.py` since they are all pointless, and installing a mail server on the vagrant box, and this appears to be working fine.  @annagav you'll want to do 

```
vagrant ssh
sudo apt-get install postfix
```

and when prompted for which type of site, pick `Internet Site`.

Let me know if you have different results, and if you prefer I can PR my changes to your branch

annagav:
It works now! Thank you!

auth/templates/uniform/allauth/account/login.html
```diff
@@ -124,7 +124,20 @@
 <form class="login uniForm" method="POST" action="{% url 'account_login' %}">
  	{% csrf_token %}
   <fieldset class="inlineLabels">
-  {{ form|as_uni_form }}
```
carsongee, line 127: if _uni_form isn't working, we should remove it from the requirements.txt as well.  You may want to try this again now that we updated it though.

requirements.txt
```diff
@@ -1,14 +1,14 @@
...
-django-filter==0.5.4
-django-tastypie==0.9.15
-django-uni-form==0.7.0
-djangorestframework==2.1.17
+django-allauth==0.19.1
+django-bootstrap-form==3.2
+django-filter==0.9.2
+django-tastypie==0.12.1
+django-uni-form==0.9.0
+djangorestframework==3.1.1
```
carsongee, line 11: while we're at it, is `django-bootstrap-form`, `django-filter`, `django-tastypie`, or `djangorestframework` even used?  I also find it very strange that we have both of the two major competing API libraries (tastypie and DRF)....sigh

annagav, line 11: yes, I thought about that. But it probably should be in a different PR.

carsongee:
Cool, well if everything is working as you expect here, I'm happy to hit the merge button.  It all looks good to me.

auth/templates/uniform/allauth/account/login.html
annagav, line 127: It is working, we just don't use it anymore :P

carsongee:
:video_camera:

PR #130, opened 2015-03-23
fixed blind indexing on group for logged user. fixes #110

### LINKED ISSUE #110: Blind indexing on group for logged in user
If the user is logged in, the call to `/scb/get_user.js` errors if the account doesn't have a group due to blindly indexing for the first group, which then causes JS errors, which causes BSOD.

See this line:
https://github.com/starteam/starcellbio_html/blob/18370f3811c45590a7938d973aa153458b58e96b/StarCellBio/views.py#L91

This is pretty low priority since this can only really happen when generating an account using a method other than the registration form (e.g. generating a service account).

annagav:


pdpinch:
:+1:

PR #134, opened 2015-03-24
Western blot -> parser_fixed 

annagav:
Modified parser_fixed to match duration as attribute in the rule, and accept drug value ANY.


pdpinch:
I don't know why github is showing this PR with 3 commits, but otherwise everything looks OK.

Issue #33, opened 2015-02-26
master_model.js needs to be modularized


bdero:
`master_model.js` (https://github.com/starteam/starcellbio_html/blob/develop/html_app/master_model.js) is a massive (16015 line) file that contains all of the assignment data. It's also the file that's hacked on the most.

It should be separated into different files; possibly one file per assignment and/or exercise.

bdero:
It's worth noting that, if/when we do this, it will take some coordination between @annagav and the person refactoring, since a large portion of her work happens in `master_model.js`.

PR #143, opened 2015-04-09
refactor assignments data

annagav:
Moving all assignments into html_app/assignments.
Moving each assignment's data into single folder with the assignment's name/code.


static_assets.yml
```diff
@@ -187,19 +187,26 @@ javascripts:
...
   - js/plugins/textcolor/plugin.min.js
   - ui/ExperimentDesignView.js
+  - assignments/scb_ex/assignment.js
+  - assignments/sutd_intro_biology_ps2/assignment.js
+  - assignments/sutd_intro_biology/assignment.js
+  - assignments/mit706s14/assignment.js
+  - assignments/mit706s14ps2/assignment.js
+  - assignments/decusability/assignment.js
+  - assignments/microscopy_usability/assignment.js
   - master_model.js
```
bdero, line 205: Shouldn't `master_model.js` also be removed from here?

bdero, line 205: Ah, nevermind. I see the file is actually still there.

bdero:
Since these change moves assignments, could you also normalize the whitespace in these files to use, say, 2 spaces per indent? Any opinions on this @carsongee ? I tend to follow the Google style guide, which does two-space indenting: https://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml#Code_formatting

I figure now's the best time to do it, since this PR will conflict with any changes to these assignments as is.

bdero:
It looks like this addresses #33.

pdpinch:
Does this leave us with the "old assignments" still in master_model.js?

Nevertheless, this seems like a big improvement. I need to test in vagrant still.

annagav:
Master_model still has OLD assignments and other function that populates some of the attributes of more recent assignments, but I haven't used it for my assignments.

pdpinch:
Which course codes do you expect to (still) work? I got sutd_biology to work, but neither of the 7.06 ones work.

Here's the list from signup.html:

```
    var code_values = ['starx', '7.02', '7.06', 'bio52', 'usability', '7.06_spring_2014', 'sutd_biology', 'scb_sampleexercises'];
```

(If you like, you can just remind me where the mapping of course codes to assignment objects lives)

annagav:
I think Lourdes and Alison would want all of them to be available for the future.
In this case '7.06' is an old version of '7.06_spring_2014', so I don't think we have to worry about it.

pdpinch:
I can't get the following course codes to work. After I create an account with the course code, and log in with the new account, I get dumped back to the sample exercises and (apparently) logged out:

starx
7.02
7.06
7.06_spring_2014
bio52 (didn't work before)
scb_sampleexercises

These ones work:

usability
sutd_biology

I can't tell if this is a regression or not, but I'd like to be reassured that shifting the assignment javascript out of master_model didn't cause these issues. 

Here are some sample error messages, which may or may not be relevant:

```
GET http://localhost:8100/static/assignments/scb_ex/gen/addmultipledialog.gss.css?_=1428955599 
localhost/:21 GET http://localhost:8100/static/test/gen/test.gss.css?_=1428955599 
localhost/:32 GET http://localhost:8100/static/assignments/mit706s13/gen/instructions.gss.css?_=1428955599 
localhost/:33 GET http://localhost:8100/static/mit706s13/gen/addmultipledialog.gss.css?_=1428955599 
localhost/:35 GET http://localhost:8100/static/assignments/tufts/gen/instructions.gss.css?_=1428955599 
```

annagav:
Out of the listed non-working courses the following should work:
-7.06_spring_2014
-scb_sampleexercises

pdpinch:
Looks like this will need to be rebased before we can think about merging it. 

@annagav I'm going to switch this back to "Waiting on Author" you can close it if you like or we can pick it up again later when the time is right.

pdpinch:
Hi Anna. I'm just checking to see if we should close this, or rebase it?

pdpinch:
Hi @annagav checking again -- close or refactor this PR? If we don't need it, you can just close it.

PR #179, opened 2015-04-24
change signup form: course_code lookup in Course model

annagav:


auth/forms.py
```diff
@@ -4,13 +4,25 @@
...
+from django.conf import settings
 
 
 class SignupForm(forms.Form):
 
+    def clean(self):
+        super(SignupForm, self).clean()
+        course_codes = self.data.get('course_code').replace(' ', '').split(';')
+        for course_code in course_codes:
+            #if the course code is empty or is neither an instructor nor a valid student course code
```
carsongee: nit: should be `# kdldl`,

auth/forms.py
```diff
@@ -4,13 +4,25 @@
...
+            if not course_code or (course_code != settings.INSTRUCTOR_CODE and
+                                           Course.objects.filter(code=course_code).count() < 1):
```
carsongee: nit: over indented, should be:

``` python
            if not course_code or (course_code != settings.INSTRUCTOR_CODE and
                                   Course.objects.filter(code=course_code).count() < 1):
```

auth/forms.py
```diff
@@ -4,13 +4,25 @@
...
+                raise forms.ValidationError('COURSE CODE is invalid', code="invalid")
+            #Later we could have the admin assign a unique code to an instructor for registering
```
carsongee: General nit, we prefer to line wrap at 80.  You can run pep8 against this to check this sort of thing.

auth/forms.py
carsongee: I think it probably makes more sense to split the instructor logic above the loop.  i.e. before anything just have:

``` python
if len(course_codes) == 1 and course_code == settings.INSTRUCTOR_CODE:
    return
```

or do we want someone to be able to have multiple course codes with an instructor code mixed in?

annagav: I was  trying not to modify the original expected behavior.

auth/forms.py
annagav: ??? :open_mouth:

annagav:
Updated!

backend/models.py
```diff
@@ -9,7 +9,7 @@
 
 
 class Course(models.Model):
-    code = models.CharField(max_length=50, primary_key=True)
+    code = models.CharField(max_length=50, primary_key=True, unique=True)
```
carsongee: If you leave this off (and remove the migration), can you double check that it does indeed enforce uniqueness?  There is an SO that suggests that you may still need to add unique=True: http://stackoverflow.com/questions/6039443/primary-key-and-unique-key-in-django

auth/forms.py
```diff
@@ -1,16 +1,46 @@
...
+    account_type = forms.ChoiceField(
+        choices=(('student', 'Student'), ('instructor', 'Instructor')),
+        required=True, label="Account Type",
+        widget=forms.RadioSelect
+    )
+
+    def clean(self):
+        super(SignupForm, self).clean()
+        form_course_codes = self.data.get('course_code').replace(' ', '')
+        log.error(form_course_codes)
```
carsongee: you should change this to log.debug, or just remove it all together

annagav:
Fixed!

auth/forms.py
```diff
@@ -5,12 +5,36 @@
 from backend.models import UserCourse, Course
 from django.contrib.auth.models import Group
 
-
 class SignupForm(forms.Form):
```
carsongee, line 9: pep8/nit two lines was correct

auth/templates/uniform/allauth/account/signup.html
```diff
@@ -104,77 +108,54 @@
...
 }
 
 
 </script>
 
+{% endblock %}
+
+
+
+
```
carsongee: nit: too many blanks

carsongee:
Other than those last tiny nits, ready to merge :bowling:

annagav:
Fixed!

PR #185, opened 2015-04-30
Switch to non-deprecated migrate command

carsongee:
Just a minor change to use migrate instead of syncdb now we are on 1.7


bdero:
:ski:

PR #186, opened 2015-04-30
Run migrations during vagrant provision

bdero:
It's worth noting that the migrations in `develop` aren't up-to-date, so this won't actually do anything until the migrations are corrected.


bdero:
Closing in favor of #185

carsongee:
lol, 2 minutes apart

carsongee:
also, they are up to date, the only ones that aren't are social, which we don't have much control over

bdero:
Yeah, but mine was ~~less correct~~ wrong lol.

PR #188, opened 2015-05-04
removing old fixtures

annagav:


carsongee:
Can you come up with a script of some sort that deletes these fixtures (just a bunch of `DELETE FROM...` SQL statements)?  New servers and developers that can just do `vagrant destroy -f; vagrant up`, but having  this bad data will be a problem on starcellbio.mit.edu and scbdev.odl.mit.edu unless we decide it is ok to delete their databases before we deploy any of the assignment builder stuff.

carsongee:
Other than that, this seems good, made both types of accounts and no 500s.

carsongee:
I'm going to go ahead and merge this since I'm about to redeploy the development server, but we will still need the SQL and/or django statements to remove the bad data at some point

annagav:
``` python
from backend.models import Assignment
id_list= ['A1', 'A1-C2', 'A2', 'A2-C2', 'A3','A3-C2']
for i in id_list: 
    Assignment.objects.filter(assignmentID=i).delete()
```

carsongee:
awesome, thanks.

PR #194, opened 2015-05-13
microscopy: eliminate slide_parser usage

annagav:


carsongee:
Excellent, the comments made me think I could actually understand the code. :hospital:

PR #241, opened 2015-06-30
ex1: no-gfp image correction. Fix #223

### LINKED ISSUE #223: NSF Exercise #1 - There should be a microscopy image for No-GFP sample
The No-GFP sample does have a microscopy image. The image is black, but it is important for students to see that the sample doesn't have any fluorescence (although we will need to tell students that the image has loaded, even though it's just black). This means that the sample needs to be able to be selected on the microscopy set up page, 'Fluorescence' should be listed underneath 'Microscopy Analysis' and then the appropriate image will load for this sample. See the Dropbox coding file for instructions on where to find the appropriate image.
https://www.dropbox.com/s/ukbilyvshuggwf7/StarCellBio%20Exercise%201%20coding%20ver%202.docx?dl=0

annagav:
#223

### DIFF
--- a/html_app/master_model.js
+++ b/html_app/master_model.js
@@ -12661,7 +12661,7 @@ var __scb_sample_1 = {
                                                 drug_list: {list: [ {drug_id: 'nc', concentration_id: '0'}]},
                                                 temperature: '0',
                                                 conditions: {
-                                                    'na': ['na']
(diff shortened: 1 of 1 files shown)


pdpinch:
Ok. I can confirm there is a black image loaded.

PR #242, opened 2015-06-30
microscopy image cursor hover is pointer

### LINKED ISSUE #225: NSF Exercises 1 & 3 - the hand cursor should appear when hovering over the microscopy image
The cursor should change to the hand cursor when the user is hovering over the microscopy image in the microscope's viewfinder. This will indicate to the user that s/he can drag the image in the viewfinder to get a better look at it.

annagav:
Fixes #225

### DIFF
--- a/html_app/ui/microscopy.gss
+++ b/html_app/ui/microscopy.gss
@@ -1304,8 +1304,8 @@ right: 15px;
 
 
 
-#scb_s_microscopy_slide_content_lens_outline > svg > image{	
-	cursor: pointer;
+.scb_s_microscopy_slide_content_lens_outline > svg > image{
+
 }
 
 
@@ -1368,6 +1368,7 @@ top: 49px;
 	bottom: -17px;
 	overflow:hidden;
 	background:black;
+	cursor: pointer;
 }
 
 


html_app/ui/microscopy.gss
```diff
@@ -1368,6 +1368,7 @@ top: 49px;
 	bottom: -17px;
 	overflow:hidden;
 	background:black;
+	cursor: pointer;
```
pdpinch, line 1371: I think `cursor: grab;` is more appropriate

annagav, line 1371: I think grab is used for text selection.

pdpinch, line 1371: That seems odd to me. Can you give me an example?

http://www.w3.org/TR/css3-ui/#cursor says `grab`:

> Indicates that something can be grabbed (dragged to be moved). Often rendered as the backside of an open hand.

If we're already using `grab` elsewhere, how about `move` ?

annagav, line 1371: Here you can see what it looks like: http://www.w3schools.com/cssref/tryit.asp?filename=trycss_cursor

pdpinch:
Ok. It doesn't work as advertised. How about 

`cursor: move` ?

annagav:
Alison confirmed that she wanted the white hand with pointed finger.

PR #243, opened 2015-07-01
light and laser for fluorescent microscopy

### LINKED ISSUE #221: NSF Exercise #1 - Laser button should be "on" and light should be "off" in microscopy experimental technique
The laser button should default to "on" and the light should be "off"/disabled for all of these samples as they are fluorescent samples and require the laser to be on and the light to be off for the images to be visible. (See Exercise #3 for how this should be). The light on/off switch should actually be disabled so the user can't turn it on and off. Currently in Exercise 3, the light switch appears as if you can turn it on/off until you click on it, then it turns to gray to indicate that the switch is disabled. The light on/off switch should be disabled from the very beginning for both exercises 1 and 3.

annagav:
Fixes #221

Light is disabled (laser is on by default) for 'antibody labeling IF' or 'fluorescence' analyses types.
Light is 'on' by default for all other samples (laser is off).

### DIFF
--- a/html_app/ui/MicroscopyView.js
+++ b/html_app/ui/MicroscopyView.js
@@ -83,8 +83,6 @@ scb.ui.static.MicroscopyView.parse = function (element) {
 
 
 scb.ui.static.MicroscopyView.scb_s_microscopy_lens_draw_slide = function(state){
-
-	
 	var model = new scb.components.ModelFactory(state.context.template);
 	model.microscopy.compute(state);
     /* state.slides contains the computed image hash */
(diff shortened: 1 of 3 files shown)


pdpinch:
Is it difficult to change the cursor for the disabled buttons? If the button is disabled/grayed-out the cursor should be the standard one. 

If this is a pain, I can merge this PR and open another, low priority issue.

annagav:
I think the 'light on' button is not working as expected. Alison said the button should be disabled  for certain microscopy analysis types, and not for others. I will fix this in a separate PR.

annagav:
In this last commit I disabled the 'Light' section under 'Microscope Controls' only for fluorescent images. The way to test this is through selecting samples that have 'antibody labeling IF' or 'fluorescence' as analysis type. For other analyses types 'Light' should be enabled.

pdpinch:
Ok, I've tested this, but I want to confirm the expected behavior. For NSF #1 and NSF #3, I can turn the laser off and on (laser off turns the image black in all cases). I cannot turn the light on or off on any of the samples in NSF #1 and NSF#3.

annagav:
The light has to disabled for 'antibody labeling IF' and 'fluorescence' analyses types (independent of the exercise number).

pdpinch:
I can't find any analysis types that _don't_ use IF or fluorescence. I looked through all the NSF exercises, the SUTD exercises (#1 isn't working for me) and 7.016.

annagav:
In microscopy usability exercise the last two samples have 3 analysis types.

pdpinch:
Thanks! I can see it working now. 

I feel like there should be a better way than all those `else if`s in MicroscopyView.js L122-152, but I can live with it.

PR #259, opened 2015-07-02
microscopy add image filters and magnitude

### LINKED ISSUE #220: NSF Exercise #1 - Microscopy (Analyze page) Objective value
Within the microscope's controls on the microscopy analyze page, we can just say "N/A" instead of "Null" next to objective. This will hold true for any image where the instructor has not entered the objective within the assignment builder.

annagav:
Fix #220

### DIFF
--- a/html_app/components/MicroscopyModelFactory.js
+++ b/html_app/components/MicroscopyModelFactory.js
@@ -69,13 +69,11 @@ scb.components.MicroscopyModelFactory = function scb_components_MicroscopyModelF
                                 if (template.slide_parser) {
                                     slide_array = template.slide_parser[collection_id][slide_type][conditions][index];
                                 } else {
-                                    if (rule.if_type){
-                                        microscopy_lane.lens_map.if_type = rule.if_type;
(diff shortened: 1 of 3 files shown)


html_app/components/MicroscopyModelFactory.js
```diff
@@ -69,13 +69,11 @@ scb.components.MicroscopyModelFactory = function scb_components_MicroscopyModelF
...
-                                    if (rule.if_type){
-                                        microscopy_lane.lens_map.if_type = rule.if_type;
-                                        slide_array=[{hash: rule.imgs_hash[index], if_type: rule.if_type}];
-                                    }else{
-                                        slide_array = [
-                                        {hash: rule.imgs_hash[index]}
-                                    ];
+                                    slide_array = rule.imgs_hash[index];
+                                    for( var x= 0; x < slide_array.length; x++){
+                                        if(!slide_array[x].mag){
```
giocalitri: please test this in the right way

giocalitri: what I meant is you should test if `slide_array[x].mag` exists and it is not `undefined`

giocalitri:
this looks good to me: merging

PR #269, opened 2015-07-14
refactor master_model

annagav:
Second attempt to refactor master_model.


pdpinch:
Functionally, this looks good. I took a quick spin through all the sample exercises and the usability course. 

What other course codes can I check?

annagav:
Also SUTD and 706 courses should work: 'SUTD_Biology' and '7.06_Spring_2014'.

pdpinch:
I was able to load all the exercises from SUTD. I had a problem choosing samples for microscopy in SUTD #1, but I presume that's a regression related to the microscopy changes. I'll check 7.06 next.

pdpinch:
7.06 works okay. It seems like some of the microscopy images aren't loading -- but that seems like a separate issue.

pdpinch:
Is there anything special I can do to test the load-order of the javascript? Just in case, I think I'll scrap my vagrant and create a new one.

pdpinch:
Sorry, it took me a little while to rebuild everything, but it all works.

:+1:

PR #280, opened 2015-07-24
facs: save scroll value for samples list

### LINKED ISSUE #172: NSF Exercise 3 - Scrolling in the samples window on the Analyze page of flow cytometry
When I have scrolled down in the samples window and select a sample (after #12), then the window automatically scrolls back up to the top sample. The scroll state should be maintained when samples are selected.

https://scbdev.odl.mit.edu/#view=facs&experiment_id=scb_Experiment_56j9bo28ea87j&assignment_id=scb_ex3&facs_id=scb_Facs_j7im5mibq9dn

annagav:
Fix #172


pdpinch:
@annagav is this ready for review? If so, can you add the "Needs Review" label?

pdpinch:
Wow, that's a lot of code just to keep it from resetting the scroll. I guess that's because you're storing the scroll value in the session state?

Anyhow, :+1:

Issue #154, opened 2015-04-14
Setup page - Add Samples pop up window should have select all and clear all buttons (NSF Exercises 2 & 3 - & all exercises)


ald32:
The pop up add samples window should have a 'SELECT ALL' and 'CLEAR ALL' buttons. The select all button appears below the checkboxes. The clear all button should be on the right side of the window, vertically aligned with the select all button. See the table in the flow cytometry technique as an example: https://scbdev.odl.mit.edu/#view=facs&experiment_id=scb_Experiment_1g0b00e19oq6im&assignment_id=scb_ex2&facs_id=scb_Facs_73pk9728768a1

ald32:
@pdpinch - This isn't marked for review, but I think it can be closed.

pdpinch:
Excellent. Thank you.

PR #287, opened 2015-07-28
add samples dialog: table scroll

### LINKED ISSUE #167: NSF Exercise 3 - Add Samples pop up window
Since this exercise has so many samples, I think we need to change how the samples show up in this small pop up window for this exercise and others like it. I think that we either need to force the scroll bar to appear in the pop up window or have only a set number of samples appear in the table so that the buttons at the bottom (cancel, add samples) always appear). Even when you scroll down to see the samples, the heading on the window should always remain, along with the X in the upper right corner. I'm leaning toward the latter option because there are just so many samples in this exercise. It's really hard to manage them all in the small pop up window.

annagav:
Fix #167 , #154

### DIFF
--- a/html_app/assignments/decusability/addmultipledialog.gss
+++ b/html_app/assignments/decusability/addmultipledialog.gss
@@ -33,7 +33,7 @@
 	font-size: 14pt;
 	color: transparent;
 	background: transparent;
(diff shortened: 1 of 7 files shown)


pdpinch:
:shipit:

PR #293, opened 2015-07-29
microscopy prep: no trash can if one condition, type

### LINKED ISSUE #289: NSF Exercise 1 - microscopy setup page
I think we can remove the trash can icons from this page as there is only one analysis/condition to select for each sample. This means that the user can either select it or unselect it (or use the select all / clear all buttons)

annagav:
Fix #289


pdpinch:
looks good

PR #294, opened 2015-07-29
wb: scroll after gel loaded

### LINKED ISSUE #292: Western blotting - scroll bar in Samples window
Can you force a scroll bar to always be present in the Samples window once the "Load gel" button is selected? A scroll bar is present when the user is adding the marker, but then it disappears once the gel is loaded.

annagav:
Fix #292


pdpinch:
IMO, this is doing the right thing, even though it's not forcing a scroll bar to always be present as Alison requested. 

:+1:

PR #295, opened 2015-07-29
facs: tab name before run samples

### LINKED ISSUE #290: NSF Exercise 3 - Change label of small tabbed window in flow cytometry
The label on the small tabbed window reads "PI" in gray when you get to the "Run" page of the flow cytometry experimental technique. Once you click "Run samples" then the label of the small tabbed window changes to what it should say, which is "EGFR A488".

annagav:
Fix #290


pdpinch:
:+1:

PR #309, opened 2015-08-10
wb: small tabs position

### LINKED ISSUE #275: Western Blot: Adjust position of tab navigation
When the user does more than 4 blots in Western Blotting, arrows appear. They should be aligned with the tabs, so there's no whitespace between the tabs and the arrow buttons. 

i.e., not like this (screenshot from NSF #2):

screen shot 2015-07-21 at 11 54 32 am

annagav:
Fix #275


pdpinch:
:smile:

PR #317, opened 2015-08-12
facs: small tab font size

### LINKED ISSUE #311: NSF Exercises - inconsistencies in font labels on small tabbed windows
The labels (when unselected and selected) on the small tabs should be consistent across all three experimental techniques. Right now, the font gets quite large in selected flow cytometry tabs. Western blotting should be used as the example that the other two techniques should follow.

annagav:
Fix #311

### DIFF
--- a/html_app/ui/facs.gss
+++ b/html_app/ui/facs.gss
@@ -1013,6 +1013,13 @@ input[type=radio]:disabled{
 display:none;
 }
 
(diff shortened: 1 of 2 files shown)


carsongee:
Took me a while to figure that out do to that style being in the wrong file. The fix looks good. :guitar:

PR #318, opened 2015-08-12
wb: more than 15 samples error

### LINKED ISSUE #310: Western blotting - Change the error message regarding 15 samples
After selecting 15 samples on the Western blotting sample prep page (in Exercise 3), I encountered an message saying that I had selected more than 15 samples and that only 15 samples are allowed (and that we recommend that you select 14 so you can add the marker). This message should only appear if more than 15 samples are selected.
In addition, the " mark should be removed at the end of the message.

annagav:
Fix #310


carsongee:
This fixes the row count issue, but the `"` still shows in the error message when you pick 16: 

image

annagav:
It should show the error if you selected more than 15 errors. Before it was showing the error if you selected more than 14 samples.

annagav:
Updated!

carsongee:
:tropical_drink:

PR #324, opened 2015-08-13
micro: two conditions per line

### LINKED ISSUE #322: Microscopy set up page: 2 conditions per line
Because this is mentioned in another issue but doesn't fit with the other issue, I'm creating a new one. As a user, it would be easier to read the microscopy setup page if I could see two conditions per line. This will avoid having to scroll too much.

annagav:
Fix #322

### DIFF
--- a/html_app/ui/microscopy.gss
+++ b/html_app/ui/microscopy.gss
@@ -90,11 +90,11 @@
 }
 
 .scb_s_microscopy_sample_checkbox{
-	width: 64px;
+    width: 53px;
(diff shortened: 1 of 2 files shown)


pdpinch:
This works, but it doesn't look great when there's only one condition, i.e. 

screen shot 2015-08-14 at 9 33 45 am

pdpinch:
I'm going to go ahead and merge this one. We can open another issue if/when we want the wrapping to be different.

PR #329, opened 2015-08-18
micro: condition length

### LINKED ISSUE #322: Microscopy set up page: 2 conditions per line

annagav:
Fix #322


pdpinch:
LGTM

screen shot 2015-08-21 at 8 19 55 am

PR #330, opened 2015-08-18
facs: single gate dragging right to left

### LINKED ISSUE #285: NSF Exercises - Flow cytometry single gate analysis tool
When I analyze the histogram with the single gate analysis tool and drag from right to left, two gates with the exact same coordinates are generated. When I use the same tool and drag from left to right, only one gate is generated, as expected.

annagav:
Fix #285


pdpinch:
:+1:

PR #335, opened 2015-08-19
wb: band intensity M2, pRAF

### LINKED ISSUE #169: NSF Exercise 3 - western blot bands are incorrect
I haven't gone through all of the bands in detail, but bands are appearing where there shouldn't be any bands. For example, the WT EGFR sample with the FLAG or His antibodies should be blank. @annagav, would you like to go through the coding Excel spreadsheet together?

annagav:
Fix #169


pdpinch:
LGTM, but one of these days you'll have to explain to me how these band intensities are coded.

PR #336, opened 2015-08-19
facs: disable add all conditions button

### LINKED ISSUE #332: 'add all' conditions button should behave like a button
> The 'Add All' button works great and as expected. I feel like there should be some way to indicate that the button doesn't work when all of the conditions have been added. I was originally envisioning a checkbox, but I'm ok with leaving it as a button. It would be great if you get the pointer cursor when hovering over the button.

button should have an appropriate cursor, and be disabled after used.

annagav:
part of #332


pdpinch:
As the name suggest, this works in facs. What about microscopy?

PR #337, opened 2015-08-20
added google analytics

pdpinch:
- added basic analytics.js code 
- added page events that fire whenever the hash changes, using a pre-existing callback
- not dealing with events that don't cause a hash change
- added analytics to a handful of static html files


pdpinch:
@bdero @carsongee what's recommended practice for configuring the Google analytics UA code?

This PR uses a production code (`UA-64448603-1`). I'd like to segregate production analytics from dev analytics. Should I set that up on the GA side using filters, or should I mint a new GA code for dev?

carsongee:
With this app it is tricky since it doesn't really use django, but in theory you would do it similar to how we did it for lore, put the code in a django setting that can be easily overridden and then render the tracking script with the code as a template context variable...I don't think index.html is rendered with context though

pdpinch:
I could continue the hackiness and render the google-analytics.js ?

I'll see what I can do on the GA side of things.

carsongee:
hehe, that would do it

Issue #333, opened 2015-08-18
Flocytometry and microscopy "Add all" functionality within the Set up pages


laleman:
When one clicks on the "Add all" underneath the "Conditions" column button within the flocytometry set up page, the radial buttons are filled from left to right and then from top to bottom, which makes seeing that all of the radial buttons have been selected more difficult. It would be good to select them from top bottom first and then left to right so that it is more visually obvious. This will also apply to the "Add all" button in the microscopy page once it becomes available.
![screen shot 2015-08-18 at 10 39 16 am](https://cloud.githubusercontent.com/assets/11359229/9339033/f0348e7e-45b5-11e5-802b-06a8b6563a93.png)
![fullsizerender](https://cloud.githubusercontent.com/assets/11359229/9339200/b4afaacc-45b6-11e5-82ed-b575d1ac0bd3.jpg)

PR #341, opened 2015-08-25
micro: add all conditions button

### LINKED ISSUE #323: Microscopy set up page: add a "select all" conditions checkbox
As a user, I would like a way to select all conditions in the microscopy setup screen, instead of having to click a radio button for each condition.

After selecting the select all checkbox, the user would see sub-rows for each of the conditions that are relevant to the sample. (As you see now once you select all of the radio buttons)

annagav:
Fix #323


pdpinch:
This is OK to merge, but do you want to also try to fix #333 at the same time?

annagav:
#333 will be a little messy I guess and it is low priority, so we don't have to fix it for the August deadline.

pdpinch:
:ok:

Issue #343, opened 2015-08-25
Western blotting exp technique - nested scroll bars in the Samples window


ald32:
Two nested scroll bars appear in the samples window when there is a sufficient number of samples to create a scroll bar for the samples list and a scroll bar for all of the items in the window (which includes the gel type, samples list, and the label 'No marker loaded'. This is not very user friendly and should be changed to only have one scroll bar in this window.

pdpinch:
can we get a screenshot?

ald32:
@pdpinch, I got a screenshot only by changing my preferences in system preferences to "always" show scrollbars. It looks a little bit different if you have the setting set to "Automatically based on mouse or trackpad", as the scrollbars are one on top of the other. 
![screen shot 2015-08-25 at 2 33 12 pm](https://cloud.githubusercontent.com/assets/1855065/9475717/74d4eb44-4b36-11e5-9ac2-59813c6918c9.png)

annagav:
This is fixed.

![screen shot 2016-10-26 at 2 16 27 pm](https://cloud.githubusercontent.com/assets/7574259/19738814/cacb0f46-9b86-11e6-89e1-7f98120090fc.png)

Issue #342, opened 2015-08-25
Western blotting exp tech - "No marker loaded" text


ald32:
The "No marker loaded" text, which appears when the user does not have a protein marker, should be always visible in the samples window. It should also be repositioned so it has a slight margin surrounding it and is not directly next to the border of the samples window.

annagav:
I think I fixed it. Does this look ok?

![screen shot 2016-10-26 at 2 16 27 pm](https://cloud.githubusercontent.com/assets/7574259/19738674/56b042de-9b86-11e6-9a1e-6fccd6f54b60.png)

annagav:
The marker stays in place when the list scrolls.

PR #349, opened 2015-08-27
wb: samples window

annagav:
#343, #342


annagav:
@pdpinch, I think I made a mistake of not creating a new branch for this PR. What should I do?

Issue #348, opened 2015-08-27
Resize the 'Confirm Setup' Dialog window


ald32:
The confirm setup dialog window is too small and is cutting off the text in the window.
![screen shot 2015-08-27 at 11 57 42 am](https://cloud.githubusercontent.com/assets/1855065/9525698/5f2e6788-4cb3-11e5-87f6-b11d576f94bb.png)

pdpinch:
I can't seem to reproduce this. Is this in Chrome? Is it specific to one experiment or another?

annagav:
To reproduce this you have to got to Exercise 3 do the experiment, select 15 samples for WB, click continue. And then, I think, this bug appears.

annagav:
fixed

PR #350, opened 2015-08-27
confirm setup window width

annagav:
#348


html_app/ui/ExperimentSetupView.js
```diff
@@ -297,7 +297,7 @@ scb.ui.static.ExperimentSetupView.scb_f_open_experiment_setup_readonly = functio
 			else{
 				$('body').prepend(scb_experiment_setup.experiment_setup_overlay());
 				$(element).attr('href', 'javascript:void(0)');
-				
+				$('#jqDialog_box').css('width', '570px');
```
pdpinch, line 300: I'm not crazy about hard-coding CSS, especially when it depends on a font rendering.

pdpinch:
I still can't reproduce the original issue, but this seems like a benign change so we'll see if it makes things better for the biologist.

PR #352, opened 2015-08-28
wb: samples list window

annagav:
#343, #342

Changed the alignment of the static list to fit the message "No Marker Added", when 15 samples added.


pdpinch:
:+1:

Issue #234, opened 2015-06-29
Assignment Setup: Pre-populate the name when the instructor chooses a template


pdpinch:

pdpinch:
closed via #362 still needs acceptance test

PR #362, opened 2015-09-10
ab: assignment default name and error

### LINKED ISSUE #235: Assignment Setup: return an error if the user doesn't set a name for the assignment

annagav:
Fix #235, #234

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -68,22 +68,38 @@ def create_new_assignment(request):
 
 
 def assignment_setup(request):
+    error = ''
     if request.method == "POST":
-        # Have to give error for empty name
-        request.session['assignment_name'] = request.POST['name']
+        assignment_name = request.POST['name']
         if request.POST['based_on']:
             request.session['based_on'] = request.POST['based_on']
-
-        if 'continue' in request.POST:
-            return redirect('common_course_setup')
+        if assignment_name:
+            request.session['assignment_name'] = assignment_name
+            if 'continue' in request.POST:
+                return redirect('common_course_setup')
+        else:  # Have to give error for empty name
+            error = "*Empty assignment name"
+    else:
+        # giving a default name to the assignment
+        assignment_name = "Assignment"
+        filtered_asgmts = models.Assignment.objects.filter(name__regex=r'Assignment.*')
+        # want to allow index one larger then there are currently assignments
+        for index in xrange(1, len(filtered_asgmts)+2):
+            if not filtered_asgmts.filter(name="Assignment {}".format(index)).count():
+                assignment_name = "Assignment {}".format(index)
+                break
 
     all_assignments = models.Assignment.objects.all()
+
     return render_to_response('instructor/assignment_setup.html',
-                              {'assignments': all_assignments,
-                               'assignment_name': request.session['assignment_name'],
-                               'based_on': request.session['based_on'],
-                               'new': request.session['new']},
-                                context_instance=RequestContext(request))
+                              {
+                                  'assignments': all_assignments,
+                                  'error': error,
+                                  'assignment_name': assignment_name,
+                                  'based_on': request.session['based_on'],
+                                  'new': request.session['new']
+                              },
+                              context_instance=RequestContext(request))
 
 @login_required
 def course_setup(request):
--- a/instructor/templates/instructor/assignments.html
+++ b/instructor/templates/instructor/assignments.html
@@ -2,12 +2,10 @@
 
 {% block content %}
     <div class='scb_s_dashboard_sidebar'>
-        <h1 class="scb_s_dashboard_sidebar_title">Your Courses:
-            <div class="scb_s_dashboard_sidebar_name scb_s_dashboard_button_wrapper"></div>
-        </h1>
+        <h1 class="scb_s_dashboard_sidebar_title">Your Assignments: </h1>
 
-        <a class='scb_s_navigation_button scb_s_dashboard_new_assignment_button'
-           href="{% url "common_create_new_assignment" %}">Create a New Assignment</a>
+        <a class='scb_ab_s_navigation_button scb_s_dashboard_new_assignment_button'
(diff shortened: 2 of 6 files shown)


instructor/common.py
```diff
@@ -68,19 +68,33 @@ def create_new_assignment(request):
...
     if request.method == "POST":
-        # Have to give error for empty name
-        request.session['assignment_name'] = request.POST['name']
+        assignment_name = request.POST['name']
         if request.POST['based_on']:
             request.session['based_on'] = request.POST['based_on']
-
-        if 'continue' in request.POST:
-            return redirect('common_course_setup')
+        if assignment_name:
```
giocalitri, line 76: question: which kind of values can `assignment_name` (or `request.POST['name']` have?

instructor/common.py
```diff
@@ -68,19 +68,33 @@ def create_new_assignment(request):
...
+            error = "*Empty assignment name"
+    else:
+        index = 1
+        found = False
+        while not found:
+            assignment_name = "Assignment" + str(index)
+            if models.Assignment.objects.filter(name=assignment_name).count() > 0:
+                index += 1
+            else:
+                found = True
```
giocalitri: this loop might be correct but is not written in a pythonic way and has a serious problem:

1-(minor) in python you do not need a guardian variable to break a loop: you can use the `break` statement: this means you can get rid of the `found` variable
2-(serious) this while is not safe: in theory it can go on forever because there is no check based on the maximum amount of cycles to run: this means you have to check the maximum amount of loops it can run. Reading the code it seems that this number is equal to `models.Assignment.objects.count()`
3-(very minor) string concatenation should not be done with the `+` operator

Now looking at the code, I think you can use a different approach: if the maximum amount of loops you can run is `models.Assignment.objects.count()`, then you can write something like:

```
for index in xrange(models.Assignment.objects.count()):  # use range if this python should run in python 3
    if models.Assignment.objects.filter(name="Assignment{0}".format(index)).count():
        break;
```

This will stop at the first `index` that gives a count > 0 and you can reuse the variable later.

This is the first step, but I'm pretty sure you might even reduce the number of queries using the right filtering, maybe filtering by the `name` starting with `"Assignment"`, sorting in ascending mode and taking the first, but I'm not sure because I do not know the data in the database.

instructor/common.py
```diff
@@ -68,19 +68,33 @@ def create_new_assignment(request):
...
     all_assignments = models.Assignment.objects.all()
+
     return render_to_response('instructor/assignment_setup.html',
                               {'assignments': all_assignments,
-                               'assignment_name': request.session['assignment_name'],
+                               'error': error,
+                               'assignment_name': assignment_name,
                                'based_on': request.session['based_on'],
                                'new': request.session['new']},
                                 context_instance=RequestContext(request))
```
giocalitri: not sure if there is any linting or pep8 check, but might want to re-indent this anyway for readability

instructor/templates/instructor/assignments.html
```diff
@@ -2,12 +2,10 @@
...
     <div class='scb_s_dashboard_sidebar'>
-        <h1 class="scb_s_dashboard_sidebar_title">Your Courses:
-            <div class="scb_s_dashboard_sidebar_name scb_s_dashboard_button_wrapper"></div>
-        </h1>
+        <h1 class="scb_s_dashboard_sidebar_title">Your Assignments: </h1>
 
-        <a class='scb_s_navigation_button scb_s_dashboard_new_assignment_button'
-           href="{% url "common_create_new_assignment" %}">Create a New Assignment</a>
+        <a class='scb_ab_s_navigation_button scb_s_dashboard_new_assignment_button'
+           href="{% url "common_create_new_assignment" %}">Create a New Assignment&nbsp; +</a>
```
giocalitri, line 8: question: why do you need the `&nbsp;`? do you need 2 spaces?

instructor/common.py
annagav, line 76: a string

instructor/templates/instructor/assignments.html
annagav, line 8: To gain more distance between Assignment and +.

giocalitri, line 8: Yes, but why the `&nbsp;` and not a CSS or another solution

instructor/common.py
```diff
@@ -68,22 +68,38 @@ def create_new_assignment(request):
...
+            if 'continue' in request.POST:
+                return redirect('common_course_setup')
+        else:  # Have to give error for empty name
+            error = "*Empty assignment name"
+    else:
+        # giving a default name to the assignment
+        assignment_name = "Assignment"
+        filtered_asgmts = models.Assignment.objects.filter(name__regex=r'Assignment.*')
+        # want to allow index one larger then there are currently assignments
+        for index in xrange(1, len(filtered_asgmts)+2):
```
giocalitri, line 87: why `+2`?

instructor/common.py
```diff
@@ -68,22 +68,38 @@ def create_new_assignment(request):
...
+            if not filtered_asgmts.filter(name="Assignment {}".format(index)).count():
+                assignment_name = "Assignment {}".format(index)
+                break
```
giocalitri, line 90: now you are looking for the first available name?

instructor/common.py
annagav, line 87: because xrange(1,n) gives a range [1, n-1]. If I already have n assignments in the database and in the worst case all of them are named "Assignment "+index then I will need the index to be able to take on the value of n+1.

instructor/templates/instructor/assignments.html
annagav, line 8: Because this is how Shloka had it in her template, and it seemed pretty straight forward. How would add space within a string using css?

instructor/common.py
annagav, line 90: I believe, before I was also looking for the first available name.

instructor/common.py
giocalitri, line 87: makes sense

giocalitri:
:shipit:  (meaning you can merge)

PR #363, opened 2015-09-11
ab: course setup show form errors

annagav:
Course setup: show form validation errors. 
Fixed a bug in providing 'based_on' assignment.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -97,19 +97,28 @@ def course_setup(request):
             # Create a new course
             user = User.objects.get(username=request.user)
             form = CourseForm(request.POST)
-            course = form.save(commit=False)
-            course.owner_id = user.id
-            course.save()
+            if form.is_valid():
+                course = form.save(commit=False)
+                course.owner_id = user.id
+                course.save()
+            else:
+                all_courses = models.Course.objects.all()
+                return render_to_response('instructor/course_setup.html',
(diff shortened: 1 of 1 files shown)


instructor/common.py
```diff
@@ -97,19 +97,28 @@ def course_setup(request):
             # Create a new course
             user = User.objects.get(username=request.user)
             form = CourseForm(request.POST)
-            course = form.save(commit=False)
-            course.owner_id = user.id
-            course.save()
+            if form.is_valid():
+                course = form.save(commit=False)
+                course.owner_id = user.id
+                course.save()
```
giocalitri, line 103: question: do you need this to be in a transaction?

instructor/common.py
```diff
@@ -97,19 +97,28 @@ def course_setup(request):
...
+            else:
+                all_courses = models.Course.objects.all()
+                return render_to_response('instructor/course_setup.html',
+                                          {'courses': all_courses,
+                                           'form': form,
+                                           'new': request.session['new']},
+                                          context_instance=RequestContext(request))
```
giocalitri, line 110: you have to indent this better

instructor/common.py
annagav, line 103: I don't think I understand the question.

giocalitri, line 103: fine

PR #364, opened 2015-09-14
ab: modify course options

annagav:
When modifying existing assignment, shows a list of courses (formset) with an extra row, where you can create a new course or change exiting ones.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -200,9 +200,7 @@ def assignment_modify(request):
 def course_modify(request):
     assignment_id = request.session['assignment_id']
     assignment = get_object_or_404(models.Assignment, pk=assignment_id)
-    CourseForm = modelform_factory(models.Course, fields=['name', 'code'])
-
-    form = CourseForm(instance=assignment.course)
+    CourseFormSet = modelformset_factory(models.Course, extra=1, fields=['name', 'code'])
     if request.method == 'POST':
         if 'course_pk' in request.POST:
             # change the course for this assignment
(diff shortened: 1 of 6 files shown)


instructor/common.py
```diff
@@ -188,19 +187,24 @@ def course_modify(request):
                 return redirect("common_assignments_edit_strains")
         else:
             # change this course's name or code
-            form = CourseForm(request.POST, instance=assignment.course)
-            if form.is_valid():
-                form.save()
+            formset = CourseFormSet(request.POST)
+            formset.clean()
```
giocalitri: question: what does `formset.clean()` do?

instructor/common.py
```diff
@@ -188,19 +187,24 @@ def course_modify(request):
...
+    formset = CourseFormSet(queryset=models.Course.objects.all())
     all_courses = models.Course.objects.all()
-    return render_to_response('instructor/course_setup.html',
+    return render_to_response('instructor/course_modify.html',
                               {'courses': all_courses,
-                               'form': form,
-                               'new':  request.session['new']},
+                               'formset': formset,
+                               },
                               context_instance=RequestContext(request))
```
giocalitri, line 232: Need to indent better

instructor/templates/instructor/course_modify.html
```diff
@@ -0,0 +1,98 @@
...
+                }
+                var url = '/ab/assignments/course_modify/';
+                $.ajax({
+                    url: url,
+                    type: 'POST',
+                    data: data
+                });
+            });
+        });
+
```
giocalitri, line 5: Question: is there a good reason to put this javascript inline in the template?

annagav:
Optimized the code a bit.

instructor/common.py
annagav: You are right. This is not needed.

instructor/templates/instructor/course_modify.html
annagav, line 5: Just for convenience. I actually think I can rewrite this functionality in much better way.

instructor/common.py
```diff
@@ -188,19 +186,24 @@ def course_modify(request):
...
-    return render_to_response('instructor/course_setup.html',
-                              {'courses': all_courses,
-                               'form': form,
-                               'new':  request.session['new']},
+    return render_to_response('instructor/course_modify.html',
+                              {
+                                  'courses': all_courses,
+                                  'formset': formset,
+                              },
                               context_instance=RequestContext(request))
```
giocalitri, line 232: nit: the right indentation is:

```
return render_to_response(
    'instructor/course_modify.html',
    {
        'courses': all_courses,
        'formset': formset,
    },
    context_instance=RequestContext(request)
)
```

PR #366, opened 2015-09-15
ab: input box when no strains

### LINKED ISSUE #238: Strains: initial view has one box and an add button
User can add more rows with the add button.

User can delete rows. If user deletes all roles, revert to the initial state of one box and an add button.

annagav:
Fix #238

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -259,7 +259,7 @@ def assignments_edit_strains(request):
     pk = request.session['assignment_id']
     assignment = get_object_or_404(models.Assignment, id=pk)
     extra_fields = 0
-    if 'add' in request.POST:
+    if 'add' in request.POST or not models.Strains.objects.filter(assignment=assignment):
         extra_fields = 1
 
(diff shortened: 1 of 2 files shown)


instructor/common.py
```diff
@@ -259,7 +259,8 @@ def assignments_edit_strains(request):
     pk = request.session['assignment_id']
     assignment = get_object_or_404(models.Assignment, id=pk)
     extra_fields = 0
-    if 'add' in request.POST:
+    strains = models.Strains.objects.filter(assignment=assignment)
+    if 'add' in request.POST or len(strains) < 1:
```
giocalitri: `models.Strains.objects.filter(assignment=assignment)` returns a list. 
in python `if []` is `false`, so you can just write something like:
`if 'add' in request.POST or models.Strains.objects.filter(assignment=assignment):`

this has also th benefit that you will do the query to the DB only if  `'add'` is not in `request.POST`

annagav: I will go with your suggestion.

giocalitri:
:shipit:  
it can be merged

PR #368, opened 2015-09-18
ab: delete strains

annagav:
Implemented deletion of strains through FormSet. When the trash button is clicked the strain is disabled and grayed out and later deleted on the submission of the formset.

### DIFF
--- a/instructor/ui/assignment_setup.css
+++ b/instructor/ui/assignment_setup.css
@@ -125,12 +125,24 @@ select:disabled {
     left: 28px;
 }
 .scb_s_delete_strain {
-    border: 0px;
+    border: 0;
     width: 24px;
     background: transparent;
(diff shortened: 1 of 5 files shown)


pdpinch:
@annagav this needs a rebase

annagav:
Rebase complete.

instructor/ui/assignment_setup.css
```diff
@@ -125,12 +125,24 @@ select:disabled {
...
     background: transparent;
     height: 20px;
     cursor: pointer;
-    top: 0px;
+    top: 0;
     position: relative;
     left: 16px;
 }
+
+.checkbox_image{
```
pdpinch: seems like this class name should be more specific to the use. How about `delete_checkbox_image` ?

PR #371, opened 2015-09-24
ab: define treatments page

annagav:
The first draft of the Treatments page. Can define name, concentration, duration, start time, temperature, collection time.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -311,30 +311,60 @@ def assignments_edit_strains(request):
                               context_instance=RequestContext(request))
 
 
-def assignments_edit_protocols(request):
+def assignments_edit_treatments(request):
     pk = request.session['assignment_id']
     assignment = get_object_or_404(models.Assignment, id=pk)
-    message = ''
-    ProtocolsFormSet = modelformset_factory(models.Protocol, extra=1, can_delete=True, exclude=['assignment'])
+    DrugFormSet = modelformset_factory(models.Drug, extra=1, can_delete=True, exclude=['assignment'])
+
+    TemperatureFormSet = modelformset_factory(models.Temperature, extra=1, can_delete=True, exclude=['assignment'])
+    CollectionTimeFormSet = modelformset_factory(models.CollectionTime, extra=1, can_delete=True, exclude=['assignment'])
+
     if request.method == "POST":
-        formset = ProtocolsFormSet(request.POST)
-        formset.clean()
-        if formset.is_valid():
-            message = "Thank you"
-            entries = formset.save(commit=False)
+        drug_formset = DrugFormSet(request.POST, prefix='drug')
+        temperature_formset = TemperatureFormSet(request.POST, prefix='temperature')
+        collection_time_formset = CollectionTimeFormSet(request.POST, prefix='collection_time')
+
+        if drug_formset.is_valid():
+            entries = drug_formset.save(commit=False)
+            for obj in drug_formset.deleted_objects:
+                obj.delete()
             for form in entries:
                 form.assignment = assignment
                 form.save()
-        else:
-            message = "Something went wrong"
+        if temperature_formset.is_valid():
+            entries = temperature_formset.save(commit=False)
+            for obj in temperature_formset.deleted_objects:
+                obj.delete()
+            for form in entries:
+                form.assignment = assignment
+                form.save()
+        if collection_time_formset.is_valid():
+            entries = collection_time_formset.save(commit=False)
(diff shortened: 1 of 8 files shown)


instructor/common.py
```diff
@@ -311,27 +311,50 @@ def assignments_edit_strains(request):
...
+            for form in entries:
+                form.assignment = assignment
+                form.save()
+        if collection_time_formset.is_valid():
+            entries = collection_time_formset.save(commit=False)
+            for obj in collection_time_formset.deleted_objects:
+                obj.delete()
+            for form in entries:
+                form.assignment = assignment
+                form.save()
```
giocalitri, line 347: I'm not sure of this code: you re-assign different values to `entries` without ever committing the `save`. Is that the right behavior you are implementing?
If so some comments about what you are doing would be helpful.

instructor/common.py
```diff
@@ -311,27 +311,50 @@ def assignments_edit_strains(request):
...
+    drug_formset = DrugFormSet(queryset=models.Drug.objects.filter(assignment=assignment), prefix='drug')
+    temperature_formset = TemperatureFormSet(queryset=models.Temperature.objects.filter(assignment=assignment),
+                                         prefix='temperature')
+    collection_time_formset = CollectionTimeFormSet(queryset=models.CollectionTime.objects.filter(assignment=assignment),
+                                         prefix='collection_time')
```
giocalitri: Nit: please indent better

instructor/common.py
```diff
@@ -311,27 +311,50 @@ def assignments_edit_strains(request):
...
 
     return render_to_response('instructor/protocols.html',
-                              {'formset': ProtocolsFormSet(
-                                  queryset=models.Protocol.objects.filter(assignment=assignment)),
-                               'message': message,
+                              {'drug_formset': drug_formset,
+                               'temperature_formset': temperature_formset,
+                               'collection_time_formset': collection_time_formset,
                                'assignment': assignment
                               },
```
giocalitri: Nit: please indent better

instructor/migrations/0002_auto_20151002_1248.py
```diff
@@ -0,0 +1,132 @@
+# -*- coding: utf-8 -*-
```
giocalitri: It would be better if you rename this migration file to something more helpful according to what the migration does.
Something like `0002_new_models.py` or even more specific

instructor/models.py
```diff
@@ -187,17 +199,17 @@ class FlowCytometrySamplePrep(models.Model):
...
-admin.site.register(Protocol)
+admin.site.register(Drug)
 admin.site.register(Strains)
-admin.site.register(StrainProtocol)
-admin.site.register(Treatments)
+admin.site.register(StrainTreatment)
+admin.site.register(Treatment)
 admin.site.register(WesternBlot)
 admin.site.register(WesternBlotAntibody)
 admin.site.register(WesternBlotAntibodyBands)
```
giocalitri, line 214: These changes raise a huge red flag: are you renaming the models or re-creating them?

Looking at the code (and at the previous migration) it looks you are creating new ones and deleting the old ones. Are you sure you do not have any data you need to preserve?

giocalitri:
I will paste a comment here, since has been outdated by a new file name, but it has not been resolved.

It would be better if you rename the migration file to something more helpful according to what the migration does.
Something like `0002_new_models.py` or even more specific

giocalitri:
Few questions and comments for this PR

instructor/common.py
annagav, line 347: I am saving each modified or created form individually, so that I can assign an assignment to each one of them.

instructor/models.py
annagav, line 214: These are new tables. The old table were not what we need. There was no data for these tables.

annagav:
Addressed the comments.

instructor/common.py
giocalitri, line 347: yes, sorry, I misread the code.

instructor/models.py
giocalitri, line 214: I am not an expert about schema migration, but it seems to me that this automatic generated migration is not complete. 
If you take a look at http://stackoverflow.com/a/12361162/257092 it seems you need to do something more than that (even if that migration is for renaming a table, not deleting it).
@ShawnMilo Do you have any suggestion?

giocalitri, line 214: Ok, after reviewing it, I think the migration is good as is

giocalitri:
Will check functionality

giocalitri:
This can be merged

PR #373, opened 2015-09-29
ab: modify course view

### LINKED ISSUE #372: Course Modify: select course checkbox with course forms.
Remove the dropdown menu for selecting a course. Add checkbox/radio button at the beginning of a line for each course form.

annagav:
Fix #372


pdpinch:
Looks good for functionality

instructor/common.py
```diff
@@ -97,66 +97,71 @@ def assignment_setup(request):
...
+            for obj in formset.deleted_objects:
+                obj.delete()
             user = User.objects.get(username=request.user)
-            form = CourseForm(request.POST)
-            if form.is_valid():
-                course = form.save(commit=False)
-                course.owner_id = user.id
+            for course in courses:
+                # need to set the owner
+                course.owner = user
```
noisecapella, line 121: The data in `request.POST` could be anything so you can't trust that it only contains the user's course information. Can you add a check that says, if course.owner exists and is some other user, raise `django.core.exceptions.PermissionDenied`? This would prevent users from modifying courses they don't own.

This check should go near the beginning of this function so it happens before anything else.

instructor/common.py
```diff
@@ -97,66 +97,71 @@ def assignment_setup(request):
...
+    CourseFormSet = modelformset_factory(models.Course, extra=1, can_delete=True, fields=['name', 'code'])
+    course_selected = 0
+    if request.method == 'POST':
+
+        # change this course's name or code
+        formset = CourseFormSet(request.POST)
+        if formset.is_valid():
+            courses = formset.save(commit=False)
+            for obj in formset.deleted_objects:
+                obj.delete()
```
noisecapella, line 117: It looks like if you delete the last course in an assignment the assignment is deleted too, probably through [delete cascading](https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.ForeignKey.on_delete). It may be simplest to just turn it off in the models but I'm not sure how that would affect everything else.

instructor/common.py
```diff
@@ -97,66 +97,71 @@ def assignment_setup(request):
...
-    CourseForm = modelform_factory(models.Course, fields=['code', 'name'])
 
-    if request.method == "POST":
-        if 'course_pk' in request.POST:
-            # Adding to an existing course
-            course = models.Course.objects.get(pk=request.POST['course_pk'])
-        else:
-            # Create a new course
+    CourseFormSet = modelformset_factory(models.Course, extra=1, can_delete=True, fields=['name', 'code'])
+    course_selected = 0
```
noisecapella: If this represents a lack of `Course` set it to `None` instead

instructor/common.py
annagav, line 117: I think we have recently decided that if there are assignments associated with this course we will not allow the instructor to delete the course (there will be a popup warning) . I haven't implemented that yet.

instructor/common.py
annagav: Ok, will do.

instructor/common.py
```diff
@@ -97,66 +97,71 @@ def assignment_setup(request):
...
-
-        if 'save' in request.POST:
-            return redirect("common_course_modify")
-        else:
-            # Continue
-            return redirect('common_assignments_edit_strains')
 
+    # if there is at least one course
+    all_courses = models.Course.objects.all()
+    if not course_selected and len(all_courses) > 0:
```
noisecapella: It looks like `course_selected` will never be set at this point so you can remove that part of the expression

annagav: You are right! I changed the code multiple times and did not notice this.

instructor/common.py
```diff
@@ -97,66 +97,71 @@ def assignment_setup(request):
...
+        course_selected = all_courses[0]
```
noisecapella, line 127: This seems unsafe, it just picks the first course in the database. It looks like the assignment should already exist anyway so it may be best just to remove this block of code.

annagav, line 127: I cannot create an assignment without a course. That's why I check to see if there is t least one course and by default want to assign the first course.

instructor/common.py
```diff
@@ -97,66 +97,71 @@ def assignment_setup(request):
...
+        create_assignment(request, course_selected)
+        if 'continue' in request.POST:
+            return redirect("common_assignments_edit_strains")
+
+    if course_selected:
```
noisecapella, line 132: At this point `course_selected` looks like it's always set to something so you may want to get rid of the check and the rest of the function after the redirect.

annagav, line 132: The variable won't be set if the instructor clicked submit on an empty form, and there are no courses created.

instructor/common.py
```diff
@@ -200,34 +207,39 @@ def assignment_modify(request):
 def course_modify(request):
     assignment_id = request.session['assignment_id']
     assignment = get_object_or_404(models.Assignment, pk=assignment_id)
-    CourseFormSet = modelformset_factory(models.Course, extra=1, fields=['name', 'code'])
+    CourseFormSet = modelformset_factory(models.Course, extra=1, can_delete=True, fields=['name', 'code'])
     if request.method == 'POST':
-        if 'course_pk' in request.POST:
+        pk = request.POST['course_pk']
+        if not long(pk) == assignment.course.id:
```
noisecapella, line 214: Please change `long` to `int` for consistency. `int` is at least as big as the standard integer type in MySQL.

instructor/common.py
```diff
@@ -200,34 +207,39 @@ def assignment_modify(request):
 def course_modify(request):
     assignment_id = request.session['assignment_id']
     assignment = get_object_or_404(models.Assignment, pk=assignment_id)
-    CourseFormSet = modelformset_factory(models.Course, extra=1, fields=['name', 'code'])
+    CourseFormSet = modelformset_factory(models.Course, extra=1, can_delete=True, fields=['name', 'code'])
     if request.method == 'POST':
-        if 'course_pk' in request.POST:
+        pk = request.POST['course_pk']
```
noisecapella, line 213: You should add a check to ensure that the course is already owned by the logged in user.

instructor/common.py
```diff
@@ -200,34 +207,39 @@ def assignment_modify(request):
...
-            new_course = models.Course.objects.get(pk=request.POST['course_pk'])
+            new_course = get_object_or_404(models.Course, pk=int(pk))
             assignment.course = new_course
             assignment.save()
+
+        # change this course's name or code
+        formset = CourseFormSet(request.POST)
+        if formset.is_valid():
+
+            instances = formset.save(commit=False)
```
noisecapella, line 224: You should add a check here to make sure that the course is either a new course or currently owned by the logged in user. You can raise `django.core.exceptions.PermissionDenied` if the check fails

instructor/common.py
```diff
@@ -97,66 +97,71 @@ def assignment_setup(request):
                                   'error': error,
                                   'assignment_name': assignment_name,
                                   'based_on': request.session['based_on'],
-                                  'new': request.session['new']
+                                  'new': request.session['new'],
                               },
                               context_instance=RequestContext(request))
 
 @login_required
 def course_setup(request):
```
noisecapella, line 106: Do we still need `course_setup`?

instructor/common.py
annagav, line 213: I think at this point we thought to allow any instructor to see all courses and assignments.

instructor/common.py
annagav, line 214: I think I checked the type of assignment.course.id and it was `long`. Is it going to convert int to a long for comparison?

instructor/ui/instructor.js
```diff
@@ -21,18 +21,22 @@ $(function(){
...
     /* If the user is choosing an existing course want
     * to pass course pk */
     $("#course_formset").submit(function(){
-        if($(".scb_ab_f_course_setup_option_choose").attr('checked') === 'checked'){
-            var course_pk = $("option:selected").val();
-            $(this).append($("<input>",{
-                    type: 'hidden',
-                    name: 'course_pk',
-                    value: course_pk
+        var course_pk = $("input:checked").attr('id');
```
noisecapella: In general you shouldn't store anything except an element id used for looking up an element in the id attribute. Instead set `data-id='...'` and use `$("input:checked").data('id')` to retrieve your value.

instructor/templates/instructor/course_modify.html
```diff
@@ -4,50 +4,44 @@
...
             {% for form in formset %}
-                <div>
-                    {{ form.name }}
-                    {{ form.code }}
-                    {{ form.id }}
-                    <div class="scb_ab_s_assignment_setup_errors">{{ form.name.errors }}</div>
-                    <div class="scb_ab_s_assignment_setup_errors">{{ form.code.errors }}</div>
+                <div class="scb_ab_s_course_form_list">
+                    <div>
+                        <input  type="radio" class="scb_ab_f_select_course" id="{{ form.id.value}}"
```
noisecapella: Use data-id instead of id as describe in my comment in instructor.js

instructor/common.py
noisecapella, line 117: Sounds good

instructor/common.py
noisecapella, line 214: You're right. Can you change `int(pk)` below to `long(pk)`?

instructor/common.py
noisecapella, line 213: A check at this point would only prevent instructors from editing courses they don't own. Are instructors allowed to edit other courses?

instructor/common.py
```diff
@@ -97,66 +97,71 @@ def assignment_setup(request):
...
+
+    if course_selected:
+        return redirect("common_course_modify")
+
+    formset = CourseFormSet(queryset=models.Course.objects.all())
+    return render_to_response('instructor/course_modify.html',
+                              {
+                                  'courses': all_courses,
+                                  'formset': formset,
+                                  'course_selected': 0,
```
noisecapella: Can you remove this line if the intent is not to select any course?

instructor/common.py
noisecapella, line 132: Makes sense

instructor/common.py
annagav, line 106: We need it for the first time to create an assignment and associate a course with it.

instructor/common.py
annagav: When this code is executed there is not a single course created so no assignment can be created yet. I will change this to  `'course_selected': course_selected` which is supposed to be None.

noisecapella: Sounds good

instructor/common.py
annagav, line 121: @pdpinch Should I create a new issue for this, and then we can discuss this with Alison?

pdpinch, line 121: Yes please

annagav, line 121: Created issue #380.

instructor/common.py
annagav, line 224: I think maybe we could display the not owned courses in a non-modifiable way, so that the instructor knows from the begging that he cannot modify courses he doesn't own, and there won't be an option for that.

noisecapella, line 224: That would be a good idea also, but we can't trust that the user hasn't modified the data being sent to the server in some way so we still need the server side check.

annagav, line 224: Ok, I'll do that.

annagav:
@noisecapella, I have made the changes. You can see them in last commit.

instructor/common.py
```diff
@@ -200,34 +208,42 @@ def assignment_modify(request):
...
+        formset = CourseFormSet(request.POST)
+        if formset.is_valid():
+
+            instances = formset.save(commit=False)
+            user = User.objects.get(username=request.user)
+            for obj in formset.deleted_objects:
+                if obj.owner == user:
+                    obj.delete()
+            for instance in instances:
+                if hasattr(instance, 'owner') and instance.owner != user:
```
noisecapella, line 230: Can you remove the `hasattr` part of the expression here? We should know for sure that `instance.owner` is a field

annagav, line 230: It turns out that when a new course form is submitted it does not have the owner attribute, since I have excluded it from the form. So 'hasattr' is checking if the course is new or not.

noisecapella, line 230: Sounds good

instructor/ui/instructor.js
```diff
@@ -21,18 +21,22 @@ $(function(){
...
     /* If the user is choosing an existing course want
     * to pass course pk */
     $("#course_formset").submit(function(){
-        if($(".scb_ab_f_course_setup_option_choose").attr('checked') === 'checked'){
-            var course_pk = $("option:selected").val();
-            $(this).append($("<input>",{
-                    type: 'hidden',
-                    name: 'course_pk',
-                    value: course_pk
+        var course_pk = $("input:checked").data('id');
```
noisecapella: I found a bug where this fails if you have marked one of the rows as deleted. It seems to happen due to a hidden checkbox Django inserts there automatically. Can you change `input:checked` to `input:checked[type=radio]` to fix this issue?

annagav:
I could not reproduce the bug, but I fixed it.

noisecapella:
Thanks! I think that's it for my comments :+1:

annagav:
Thank you!

PR #403, opened 2015-10-27
ab: define vars, minor edits

### LINKED ISSUE #399: Experiment setup: Minor edits on the 'Define variables' page
- The titles above the text boxes should be in bold and made slightly bigger.
- 'Name' should be changed to 'Treatment Name'.
- The units column should be labeled. How much room do we have to write 'Concen. Units', 'Duration Units', etc?

annagav:
Fix #399

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -352,23 +352,24 @@ def assignments_edit_treatments(request):
     has_start_time = assignment.has_start_time
     has_duration = assignment.has_duration
     drug_formset_exclude = ['assignment']
+    # For Drug form
     input_headers = []
 
     if not assignment.has_concentration:
         drug_formset_exclude.extend(['concentration', 'concentration_unit'])
     else:
-        input_headers.extend(['Concent.', 'Units'])
+        input_headers.extend(['Concen.', 'Concen. Units'])
 
     if not (has_duration or has_start_time):
         drug_formset_exclude.extend(['duration', 'start_time', 'time_unit'])
     elif not has_duration:
         drug_formset_exclude.append('duration')
-        input_headers.extend(['Start Time', 'Units'])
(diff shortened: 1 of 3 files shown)


instructor/common.py
```diff
@@ -345,23 +345,24 @@ def assignments_edit_treatments(request):
...
     if not (has_duration or has_start_time):
         drug_formset_exclude.extend(['duration', 'start_time', 'time_unit'])
     elif not has_duration:
         drug_formset_exclude.append('duration')
-        input_headers.extend(['Start Time', 'Units'])
+        input_headers.extend(['Start Time', 'Time Units'])
     elif not has_start_time:
         drug_formset_exclude.append('start_time')
-        input_headers.extend(['Duration', 'Units'])
+        input_headers.extend(['Duration', 'Duration Units'])
```
noisecapella, line 370: It seems like "Time Units" would be more consistent here but on the other hand the issue specifically says to use "Duration Units", so I'm fine with this

instructor/ui/assignment_setup.css
```diff
@@ -176,9 +176,11 @@ select:disabled {
 .scb_ab_s_input_headers>div{
     width: 155px;
     display: inline-block;
+    /*font-weight: bold;*/
```
noisecapella: Stray comment

noisecapella:
The headers don't line up with the textboxes. Maybe a `<table>` would work better for that? This is outside the scope of this PR however, it's like that on develop too

instructor/ui/assignment_setup.css
noisecapella: Actually it looks like this should be uncommented since the issue requires a bold title

noisecapella:
Just the note on the commented out CSS style, after that's fixed feel free to merge

PR #405, opened 2015-10-28
ab: update samples when variables change

annagav:
When Strains or Treatments are created, update strain treatment protocols. Instead of deleting samples, the user can uncheck the 'enabled' checkbox.
Edit: 
When reselecting experimental variables:
1) Show warning only once
2) delete all samples, and create new samples. 
Order samples by Strain, Treatment name, Temperature, Collection time.
Added a button to hide unselected samples.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -265,18 +265,18 @@ def assignments_variables(request):
     if request.method == "POST":
         form = AssignmentForm(request.POST, instance=assignment)
         if form.is_valid():
-            # If form has changed want to delete existing Treatments
-            if form.has_changed() and treatments_created:
-                models.StrainTreatment.objects.filter(assignment=assignment).delete()
+            if form.has_changed():
                 models.Treatment.objects.filter(assignment=assignment).delete()
-            # Want to save the form only if at most 3 vars are selected
-            form.save(commit=False)
-            num_variables = 0
-            for field in var_fields:
-                if getattr(assignment, field):
-                    num_variables += 1
-            if num_variables <= max_num_of_vars:
-                form.save()
+                models.StrainTreatment.objects.filter(assignment=assignment).delete()
+                create_treatments(assignment)
+                # Want to save the form only if at most 3 vars are selected
+                form.save(commit=False)
+                num_variables = 0
+                for field in var_fields:
+                    if getattr(assignment, field):
+                        num_variables += 1
+                if num_variables <= max_num_of_vars:
+                    form.save()
             if 'continue' in request.POST:
                 return redirect("common_assignments_edit_treatments")
     # Refresh the assignment from the database
@@ -334,9 +334,13 @@ def assignments_edit_strains(request):
             entries = formset.save(commit=False)
             for obj in formset.deleted_objects:
                 obj.delete()
-            for form in entries:
-                form.assignment = assignment
-                form.save()
+            for strain in entries:
+                if strain.id is None:
+                    strain.assignment = assignment
+                    strain.save()
+                    create_strain_treatments(assignment, strains=[strain])
+                else:
+                    strain.save()
 
         if 'continue' in request.POST:
             return redirect('common_assignments_variables')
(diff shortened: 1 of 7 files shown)


noisecapella:
Can you rebase this please?

annagav:
Done

instructor/common.py
```diff
@@ -370,10 +370,29 @@ def assignments_edit_treatments(request):
...
         input_headers.extend(['Start Time', 'Duration', 'Time Units'])
-
-    DrugFormSet = modelformset_factory(models.Drug, extra=1, can_delete=True, exclude=drug_formset_exclude)
-    TemperatureFormSet = modelformset_factory(models.Temperature, extra=1, can_delete=True, exclude=['assignment'])
-    CollectionTimeFormSet = modelformset_factory(models.CollectionTime, extra=1, can_delete=True, exclude=['assignment'])
+    # If instructor clicked ADD, add extra form
+    drug_extra_form = 0
+    temperature_extra_form = 0
+    collection_extra_form = 0
+    if 'add_drug' in request.POST or not models.Drug.objects.filter(assignment=assignment):
```
noisecapella: It's more efficient to add `.exists()` at the end of the QuerySet because it will check for existence in SQL rather than in Python. The way it is right now it will get the rows from the database and then apply `not` to it to convert to a boolean, with zero rows being a False and more than zero being true.

If you're ever curious about the SQL being generated you can turn on SQL logging in `settings.py`, run `./manage.py shell` and try out different QuerySets to see what they produce. To turn on SQL logging I made this adjustment to `LOGGING` in `settings.py`:

```
LOGGING = {
    'version': 1,
    'disable_existing_loggers': False,
    'filters': {
        'require_debug_false': {
            '()': 'django.utils.log.RequireDebugFalse'
        }
    },
    'handlers': {
        'mail_admins': {
            'level': 'ERROR',
            'filters': ['require_debug_false'],
            'class': 'django.utils.log.AdminEmailHandler'
        },
        'console': {
            'level': 'DEBUG',
            'class': 'logging.StreamHandler',
            'formatter': 'verbose'
        }
    },
    'formatters': {
        'verbose': {
            'format': (
                '[%(asctime)s] %(levelname)s %(process)d [%(name)s] '
                '%(filename)s:%(lineno)d - '
                '[{hostname}] - %(message)s'
            ).format(hostname="VAGRANT"),
            'datefmt': '%Y-%m-%d %H:%M:%S'
        }
    },
    'loggers': {
        '': {
            'handlers': ['console'],
            'level': 'DEBUG'
        },
        'django.request': {
            'handlers': ['mail_admins'],
            'level': 'ERROR',
            'propagate': True,
        },
    }
}
```

This logs absolutely everything to console which may be a bit much but can be helpful in development.

instructor/common.py
```diff
@@ -482,10 +505,13 @@ def strain_treatments_edit(request):
                               context_instance=RequestContext(request))
 
 
-def create_treatments(assignment):
-    drugs = models.Drug.objects.filter(assignment=assignment)
-    temperature = models.Temperature.objects.filter(assignment=assignment)
-    collection_times = models.CollectionTime.objects.filter(assignment=assignment)
+def create_treatments(assignment, drugs=None, temperature=None, collection_times=None):
```
noisecapella: Can you rename `temperature` to `temperatures` for consistency with the other arguments?

instructor/common.py
```diff
@@ -383,23 +402,35 @@ def assignments_edit_treatments(request):
             entries = drug_formset.save(commit=False)
             for obj in drug_formset.deleted_objects:
                 obj.delete()
-            for form in entries:
-                form.assignment = assignment
-                form.save()
+            for drug_instance in entries:
```
justinabrahms: It looks like we have almost identical code repeated three times. If we had an object that mapped the form prefix to a tuple of formset and kwarg name to create_treatments, we could DRY it up.

``` python
mapping = {'drug': (DrugFormSet, 'drugs'), ...}
for prefix, (ModelForm, treatment_kw) in mapping.items():
  formset = ModelForm(request.POST, prefix=prefix)
  if formset.is_valid():
    entries = formset.save(commit=False)
    for obj in formset.deleted_objects:
      obj.delete()
    for instance in entries:
      if instance.id is None:
        instance.assignment = assignment
        instance.save()
        create_treatments(assignment, **{treatment_kw: instance})
      else:
        instance.save()
```

instructor/common.py
```diff
@@ -261,10 +261,6 @@ def assignments_variables(request):
     if request.method == "POST":
         form = AssignmentForm(request.POST, instance=assignment)
         if form.is_valid():
-            # If form has changed want to delete existing Treatments
```
noisecapella, line 268: If I uncheck something like 'Collection Time' for example, and I previously had two collection times defined, the number of treatment combinations in `edit_strain_treatments` should shrink by half since there aren't two collection times anymore. Is there a way to do this easily?

noisecapella:
I like the UI, it feels intuitive to me!

instructor/common.py
annagav, line 268: This is exactly what we discussed with Alison today. Do you think deleting and recreating all objects is a good idea? That would be the easiest :). Otherwise, I could set the corresponding field to null and  remove all duplicates. But this would work only if I am unselecting variables.

noisecapella, line 268: I think it would be fine to delete and recreate the objects, the only data being lost is which strain treatment combinations the user selected.

instructor/common.py
```diff
@@ -265,18 +265,18 @@ def assignments_variables(request):
...
-            # Want to save the form only if at most 3 vars are selected
-            form.save(commit=False)
-            num_variables = 0
-            for field in var_fields:
-                if getattr(assignment, field):
-                    num_variables += 1
-            if num_variables <= max_num_of_vars:
-                form.save()
+            if form.has_changed():
+                models.Treatment.objects.all().delete()
```
noisecapella: Wouldn't you still need to filter this on assignments?

annagav: yes :D

annagav:
Fixed

instructor/migrations/0005_auto_20151030_1112.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
noisecapella: Would it be possible to rename this migration? Something like 0005_removed_order_field would work

instructor/ui/instructor.js
```diff
@@ -144,6 +146,16 @@ $(function(){
...
+    $(".toggle_protocols").click(function(){
+        var show= $(".toggle_protocols").data("show");
+        if(show){
+            $(".scb_ab_s_row").show();
+            $(".toggle_protocols").data("show", false).html("SHOW ONLY SELECTED");
+        }else{
+            $("input:checkbox:not(:checked)").parent().parent().hide();
+            $(".toggle_protocols").data("show", true).html("SHOW ALL");
+        }
+    })
```
noisecapella: Missing semicolon

instructor/ui/instructor.js
```diff
@@ -107,12 +107,14 @@ $(function(){
             );
             $('.jqDialog_header').remove();
             $('#jqDialog_box').prepend("<h1 class='jqDialog_header' role='heading' >Warning</h1>");
+            showed_warning=true;
```
noisecapella, line 110: This will stop showing the warning whether the user clicked yes or no, but if that's your intention it's fine

annagav, line 110: As a user, would you have expected it to continue showing the message if you clicked 'no'?

instructor/common.py
```diff
@@ -459,24 +477,23 @@ def strain_treatments_edit(request):
...
-        if not models.Treatment.objects.filter(assignment=assignment).exists():
-            create_treatments(assignment)
-            create_strain_treatments(assignment)
 
-    formset = STFormSet(queryset=models.StrainTreatment.objects.filter(assignment=assignment))
+    formset = STFormSet(
+        queryset=models.StrainTreatment.objects.filter(assignment=assignment).order_by(
+            'strain',
+            'treatment__drug__name',
+            'treatment__temperature__degrees',
```
noisecapella, line 493: Since these fields are text this will be ordered in alphabetical order, just FYI. It's fine as is

instructor/ui/instructor.js
noisecapella, line 110: You're right, the current behavior seems reasonable to me

instructor/common.py
annagav, line 493: That's right. But it shouldn't be a problem if temperatures are below 100. I am not sure if in the future this value will be ever used/needed as a number.

noisecapella:
Everything looks good :+1:

annagav:
Thank you! :dancer:

PR #417, opened 2015-11-03
ab: treatments, extra fields logic

annagav:
Need extra fields to be computed after the changes in the POST request have been saved.


noisecapella:
:+1:

PR #418, opened 2015-11-03
ab: wb antibodies page

annagav:


instructor/common.py
```diff
@@ -583,32 +583,40 @@ def western_blot_lysate_type(request):
...
-                              },
-                              context_instance=RequestContext(request))
+    formset = WesternBlotAntibodyFormset(
+        queryset=models.WesternBlotAntibody.objects.filter(western_blot=wb))
+    return render_to_response(
+        'instructor/wb_antibody.html',
+        {
+            'formset': formset,
+            'assignment_name': assignment.name,
+            'section_name': 'Experiment Setup'
```
noisecapella: This doesn't match the header but feel free to leave it if the intent is different

instructor/common.py
```diff
@@ -583,32 +583,40 @@ def western_blot_lysate_type(request):
...
 
-def western_blot_antibody_edit(request, assignment):
-    a = models.Assignment.objects.get(id=assignment)
-    wb, created = models.WesternBlot.objects.get_or_create(assignment=a)
-    wb.save()
-    WesternBlotAntibodyFormset = modelformset_factory(models.WesternBlotAntibody, extra=1, can_delete=True,
-                                                      exclude=['western_blot'])
-    message = ''
+    extra_fields = 0
+    if 'add' in request.POST or not models.WesternBlotAntibody.objects.filter(western_blot=wb):
```
noisecapella: If you have something in the last row and you delete it, no fields show up. This isn't terribly important since the user can just click 'Add' to bring back the field, but it differs in behavior from the strains treatments page. If you calculate `extra_fields` after the delete I think you would avoid this problem

annagav: Fixed!

instructor/common.py
annagav: Fixed!

noisecapella:
:+1:

PR #423, opened 2015-11-12
ab: wb intensity bands page

### LINKED ISSUE #416: Western blot: set intensity for each treatment/antibody combination
This is a very long table with sliders for setting the intensity. 

The intensity slider will have 4 snap-to settings: 0 (no display), low, medium and high

annagav:
Fix #416, #424

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -656,11 +656,83 @@ def western_blot_band_size(request):
             entries = formset.save(commit=False)
             for form in entries:
                 form.save()
+            if 'continue' in request.POST:
+                return redirect('western_blot_band_intensity')
     formset = AntibodiesFormset(queryset=models.WesternBlotAntibody.objects.filter(western_blot=wb))
     return render_to_response(
         'instructor/wb_band_size.html',
         {
             'formset': formset,
+            'assignment_name': assignment.name,
+            'section_name': 'Western Blotting'
+        },
+        context_instance=RequestContext(request))
+
+
+def is_float(txt):
+    try:
+        float(txt)
+        return True
+    except ValueError:
+        return False
+
+@login_required
+def western_blot_band_intensity(request):
+    pk = request.session['assignment_id']
+    assignment = models.Assignment.objects.get(id=pk)
+    wb, created = models.WesternBlot.objects.get_or_create(assignment=assignment)
+    BandsFormset = modelformset_factory(
+        models.WesternBlotBands,
+        extra=0,
+        exclude=['strain_protocol', 'antibody', 'weight', 'lysate_type']
+    )
+    # Create Bands
+    antibodies = models.WesternBlotAntibody.objects.filter(western_blot=wb)
+    strain_treatments = models.StrainTreatment.objects.filter(assignment=assignment)
+    weights_by_type = ['wc_weight', 'nuc_weight', 'cyto_weight']
+    for antibody in antibodies:
+        for strain_treatment in strain_treatments:
+            for index, field in enumerate(weights_by_type):
+                weight_str = getattr(antibody, field)
+                weights = [float(s) for s in weight_str.split(',')
+                           if is_float(s) and float(s) >= 0]
+                for weight in weights:
+                    models.WesternBlotBands.objects.get_or_create(
+                        strain_protocol=strain_treatment,
+                        antibody=antibody,
+                        weight=weight,
+                        lysate_type=field.split('_')[0]
+                    )
+    if request.method == 'POST':
+        formset = BandsFormset(request.POST)
+        if formset.is_valid():
+            entries = formset.save(commit=False)
+            for entry in entries:
+                entry.save()
+    else:
+        formset = BandsFormset(queryset=models.WesternBlotBands.objects.filter(antibody__western_blot=wb))
+
+    formset_group = []
+    for antibody in antibodies:
+        form_list = [form for form in formset
+                     if form.instance.antibody.primary == antibody.primary]
+        formset_group.append((antibody, form_list))
+
+    variables = {
+        'has_concentration': assignment.has_concentration,
+        'has_start_time': assignment.has_start_time,
+        'has_duration': assignment.has_duration,
+        'has_temperature': assignment.has_temperature,
+        'has_collection_time': assignment.has_collection_time
+    }
+
+    return render_to_response(
+        'instructor/wb_band_intensity.html',
+        {
+            'formset': formset,
+            'formset_group': formset_group,
+            'antibodies': antibodies,
+            'variables': variables,
             'assignment_name': assignment.name,
             'section_name': 'Western Blotting'
         },
--- a/instructor/templates/instructor/wb_band_intensity.html
+++ b/instructor/templates/instructor/wb_band_intensity.html
@@ -1,0 +1,78 @@
+{% extends "instructor/assignment_sidebar.html" %}
+
+{% block right_side %}
+    <div class="scb_ab_s_heading_div">
+        <div class="scb_ab_s_page_title">Band Intensity</div>
+        <div class='scb_ab_s_heading_description'>
+            Define which bands will appear and their relative intensity for each of your samples in your experiment setup.
+        </div>
+
+
+    </div>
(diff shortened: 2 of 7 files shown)


instructor/ui/instructor.js
```diff
@@ -163,6 +163,30 @@ $(function(){
     $("#id_has_nuclear_fractination").attr('disabled', true).parent().addClass('scb_ab_s_grayed');
     $("#id_has_cytoplasmic_fractination").attr('disabled', true).parent().addClass('scb_ab_s_grayed');
 
+    /**
+     * Western Blotting: band intensity
+     */
+    var sliders = $(".intensity_slider");
+    for(var i=0; i<sliders.length;i++){
```
noisecapella: It may be cleaner to use `$(".intensity_slider").each(function($slider) { ...` instead, since you're not using the `i` variable anywhere else.

instructor/ui/instructor.js
```diff
@@ -163,6 +163,30 @@ $(function(){
...
+        var $slider=$(sliders[i]);
+        var intensity_dict={0:'', 0.5:'low', 1:'med', 1.5:'high', 2:'extra high'};
+        var intensity=$($slider.siblings('input')[0]).attr('value');
+        $($slider.parent().siblings('.scb_ab_s_intensity_text')).html(intensity_dict[intensity]);
```
noisecapella: It's safer to use `.text(intensity_dict...` instead of `.html` if you aren't intending to put any HTML in the element.

instructor/ui/instructor.js
```diff
@@ -163,6 +163,30 @@ $(function(){
     $("#id_has_nuclear_fractination").attr('disabled', true).parent().addClass('scb_ab_s_grayed');
     $("#id_has_cytoplasmic_fractination").attr('disabled', true).parent().addClass('scb_ab_s_grayed');
 
+    /**
+     * Western Blotting: band intensity
+     */
+    var sliders = $(".intensity_slider");
+    for(var i=0; i<sliders.length;i++){
+        var $slider=$(sliders[i]);
+        var intensity_dict={0:'', 0.5:'low', 1:'med', 1.5:'high', 2:'extra high'};
```
noisecapella: This will work fine but be careful about using floating point numbers as keys in a hashtable: http://stackoverflow.com/questions/22373549/is-it-okay-to-use-floating-point-numbers-as-keys-in-objects

instructor/ui/instructor.js
```diff
@@ -163,6 +163,30 @@ $(function(){
...
+        $($slider.parent().siblings('.scb_ab_s_intensity_text')).html(intensity_dict[intensity]);
+        $slider.slider({
+          min: 0,
+          max: 2,
+          range: "min",
+          value: intensity,
+          step: 0.5,
+          slide: function( event, ui ) {
+              $(this).siblings()[0].value=ui.value;
+              $($(this).parent().siblings('.scb_ab_s_intensity_text')).html(intensity_dict[ui.value]);
```
noisecapella: It looks like this is the same selection as line 174, is there some way to store it in a variable to reuse it?

instructor/ui/instructor.js
noisecapella: Actually it looks like there is a floating point comparison problem. If you set the sliders to different levels, save it, go to a different page and come back, some of the text shows up and some doesn't. When the floating point keys are put into an object they're converted to strings, so the `2` is actually stored as `"2"`. In the `value` attribute it's being stored as a `"2.0"` so the two strings don't compare as being equal.

It's probably best to just use a series of `if` statements, something like

```
var intensity = parseFloat(intensity);
if (intensity >=0 && intensity < 0.75) ...
else if (intensity >= 0.75 && intensity < 1.25) ...
```

to avoid the problems with floating point comparison

instructor/templates/instructor/wb_band_intensity.html
```diff
@@ -0,0 +1,73 @@
...
+            </div>
+
+            {% for form in formset %}
+                <div class="scb_ab_s_row">
+                    <div class="scb_ab_s_col_width_sample">
+                        {{ forloop.counter }}.
+                        {{ form.instance.strain_protocol.strain.name }},
+                        {{ form.instance.strain_protocol.treatment.drug.name }}
+                        {% if variables.has_concentration %}
+                            ,{{ form.instance.strain_protocol.treatment.drug.concentration }}
```
noisecapella: The comma is being written against the right side instead of the left, ie `a ,b ,c`. If you add a space after the comma here and remove all space before it, it should look more like `a, b, c`

noisecapella:
The 'Save and Continue' button for `wb_band_size` doesn't proceed to the next page

instructor/common.py
```diff
@@ -667,6 +667,61 @@ def western_blot_band_size(request):
...
+    )
+    # Create Bands
+    antibodies = models.WesternBlotAntibody.objects.filter(western_blot=wb)
+    strain_treatments = models.StrainTreatment.objects.filter(assignment=assignment)
+    weights_by_type = ['wc_weight', 'nuc_weight', 'cyto_weight']
+    for antibody in antibodies:
+        for strain_treatment in strain_treatments:
+            for index, field in enumerate(weights_by_type):
+                weight_str = getattr(antibody, field)
+                weights = [int(s) for s in weight_str.split(',') if s.isdigit()]
```
noisecapella: If there's a space in `s` then `isdigit` will reject it. Can you add a `.strip()` to remove whitespace?

instructor/templates/instructor/wb_band_intensity.html
annagav: The space before the comma appears because each variable is on a new line. I haven't found a way to get rid of the space without putting everything on one line.

instructor/common.py
annagav: I will change this to recognize float numbers too.

annagav:
Updated!

instructor/ui/instructor.js
```diff
@@ -163,6 +163,44 @@ $(function(){
...
+        if(num<0.25){
+            return '';
+        }else if(num<0.75){
+            return "low";
+        }else if(num<1.25){
+            return 'med';
+        }else if(num<1.75){
+            return 'high';
+        }else{
+            return 'extra high'
```
noisecapella: Can you add a semicolon to this line?

instructor/templates/instructor/wb_band_intensity.html
noisecapella: You wouldn't need to put everything on one line, just the things before and after the comma.

```
    {{ form.instance.strain_protocol.strain.name }},
    {{ form.instance.strain_protocol.treatment.drug.name }}{% if variables.has_concentration %},
      {{ form.instance.strain_protocol.treatment.drug.concentration }}
        {{ form.instance.strain_protocol.treatment.drug.concentration_unit }}
    {% endif %}{% if variables.has_start_time %},
```

instructor/common.py
```diff
@@ -667,6 +669,81 @@ def western_blot_band_size(request):
...
+
+
+
+    return render_to_response(
+        'instructor/wb_band_intensity.html',
+        {
+            'formset': formset,
+            'formset_group': formset_group,
+            'antibodies': antibodies,
+            'number': 0,
```
noisecapella: Can you remove this since it isn't being used in the template?

instructor/common.py
```diff
@@ -667,6 +669,81 @@ def western_blot_band_size(request):
...
+        if formset.is_valid():
+            entries = formset.save(commit=False)
+            for entry in entries:
+                entry.save()
+    else:
+        formset = BandsFormset(queryset=models.WesternBlotBands.objects.filter(antibody__western_blot=wb))
+
+    formset_group = []
+    for antibody in antibodies:
+        form_list = []
```
noisecapella: This can be rewritten as a list comprehension: 

```
form_list = [form for form in form_set 
 if form.instance.antibody.primary == antibody.primary]
```

annagav:
Done!

noisecapella:
:+1:

PR #437, opened 2015-11-24
ab: error for band size values

### LINKED ISSUE #426: Western blot: error message for non-numerical values on band size page
If the instructor enters non-numerical values or negative numbers for band sizes, then the following error message should appear:
You have entered non-numerical values, including negative numbers, text inputs and/or symbols, in the band size input boxes. The band size input boxes must only contain numerical values before you may continue. 

User must click 'OK'

annagav:
Fix #426

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -12,6 +12,7 @@ from django.contrib.auth.decorators import login_required
 from django.shortcuts import get_object_or_404
 from django.core.exceptions import PermissionDenied
 import json
+import re
 
 @login_required
 def courses(request):
@@ -669,6 +670,12 @@ def western_blot_band_size(request):
     pk = request.session['assignment_id']
     assignment = models.Assignment.objects.get(id=pk)
     wb, created = models.WesternBlot.objects.get_or_create(assignment=assignment)
+    error = ''
+    message = "You have entered non-numerical values, " \
+              "including negative numbers, text inputs and/or symbols, " \
+              "in the band size input boxes. The band size input boxes must " \
+              "only contain numerical values."
+    
     exclude_fields = ['primary', 'secondary', 'western_blot']
     lysate_types = {
         'has_whole_cell_lysate': 'wc_weight',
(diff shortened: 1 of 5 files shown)


noisecapella:
It looks like the error message is shown after the save has already been executed. Is the intention just to warn the user that their input is bad or should the application prevent the user from submitting this input in the first place, like the application does with the checkboxes in other places?

annagav:
I think we decided to just warn the user and let him change it. No bands will be generated from invalid input.

instructor/templates/instructor/wb_band_size.html
```diff
@@ -1,6 +1,9 @@
 {% extends "instructor/assignment_sidebar.html" %}
 
 {% block right_side %}
+    <script>
+    var num_error = {{ num_error }}
```
noisecapella: Can you add a semicolon and also change the template variable to `{{ num_error|safe }}`? It's not strictly necessary here since `num_error` is always a number but it's good practice when JSON encoding input to mark it safe so it's not escaped in the template.

instructor/common.py
```diff
@@ -723,6 +729,7 @@ def western_blot_band_size(request):
             'formset': formset,
             'types_selected': types_selected,
             'assignment_name': assignment.name,
+            'num_error': json.dumps(error),
```
noisecapella: Can you rename num_error to has_error or something to that effect, since it's a bool?

instructor/common.py
```diff
@@ -694,6 +698,8 @@ def western_blot_band_size(request):
             for antibody in antibodies:
                 for index, field in enumerate(weights_by_type):
                     weight_str = getattr(antibody, field)
+                    error = True if bool(re.search(r'[a-zA-Z-]+', weight_str)) else error
```
noisecapella: `bool()` isn't necessary here since the `if` expression will do that evaluation for you

instructor/common.py
noisecapella: Actually you may want to create the error message here instead and send it to the javascript to be shown. It makes the code in `instructor.js` less tied to this particular view

annagav:
fixed!

noisecapella:
:+1:

PR #442, opened 2015-11-30
ab: update compiler.py

annagav:


pdpinch:
I'm not sure how we should review this?

annagav:
Now you should be able to click on preview in the dashboard and see the assignments page.

ald32:
@annagav , can I see this yet?

annagav:
@ald32, you cannot see this yet, it is still under review.

instructor/compiler.py
```diff
@@ -347,17 +337,17 @@ def primary_anti_body(assignment):
         for band in a.bands.all():
             if western_blot.has_whole_cell_lysate:
```
noisecapella: Is it possible to combine these if statements? Is it possible for `western_blot` to have more than one `has_` flag set to true at a time?

html_app/assignments/scb_ex/addmutlpledialog.soy
```diff
@@ -1,93 +0,0 @@
-{namespace scb_ex_old}
-
-/**
-@param template
-@param assignment
-@param experiment
-*/
-{template .dialog}
-<div class='scb_ex_inner_dialog' role='dialog' aria-label='Add Samples'>
```
noisecapella, line 9: Is there an easy way to view this dialog box somewhere? I'm assuming this was a duplicate of the `{template .dialog}` in the file above but there are minor differences

instructor/compiler.py
```diff
@@ -372,56 +362,49 @@ def primary_anti_body(assignment):
...
-        protocol = strain_protocol.protocol
-        for t in protocol.treatments.all():
-            tr = t.concentration
-            ret[str(tr)] = {
-                'name': str(tr),
-                'value': tr
-            }
+    for strain_protocol in assignment.strain_treatment.filter(enabled=True):
+        treatment = strain_protocol.treatment
+        tr = treatment.drug.concentration
```
noisecapella: Can you rename `tr` to `concentration`?

instructor/compiler.py
```diff
@@ -372,56 +362,49 @@ def primary_anti_body(assignment):
...
+        ret[str(tr)] = {
```
noisecapella: It seems like `str()` is unnecessary here since `concentration` is a `CharField`

instructor/compiler.py
```diff
@@ -372,56 +362,49 @@ def primary_anti_body(assignment):
...
-        protocol = strain_protocol.protocol
-        for t in protocol.treatments.all():
-            tr = t.temperature
-            ret[str(tr)] = {
-                'name': str(tr),
-                'value': tr
-            }
+    for strain_protocol in assignment.strain_treatment.filter(enabled=True):
+        treatment = strain_protocol.treatment
+        t = treatment.temperature
```
noisecapella: Can you rename `t` and `tr` to `temperature` and `degrees` respectively?

instructor/compiler.py
annagav: I honestly think this 'marks' list is not needed at all. It is useful only if I want to set a band for this particular antibody to appear for all samples. I don't think such case is common, and the assignment builder does not provide such option to the instructor.

html_app/assignments/scb_ex/addmutlpledialog.soy
annagav, line 9: If you change the namespace here to scb_ex and in addmultipledioalog2.soy change the namespace to something else, and go to setup any of the NSF sample exercises, this older version of the dialog will be used.

noisecapella:
:+1:

PR #454, opened 2015-11-30
ab: wb_intensity sample ordering

annagav:


instructor/common.py
```diff
@@ -765,7 +765,13 @@ def western_blot_band_intensity(request):
                 entry.save()
     else:
         formset = BandsFormset(queryset=models.WesternBlotBands.objects.filter(
-            antibody__western_blot=wb).order_by('strain_protocol', '-lysate_type'))
+            antibody__western_blot=wb).order_by(
+            'strain_protocol__strain',
+            'strain_protocol__treatment__drug__name',
+            'strain_protocol__treatment__temperature__degrees',
+            'strain_protocol__treatment__collection_time__time',
```
noisecapella, line 772: The order here is a little tricky since some of these values are numerical and 'time' doesn't have any units, but feel free to leave it as is since it's not an easy problem to solve

noisecapella:
:+1:

PR #455, opened 2015-12-01
ab: make new assignments private

annagav:


instructor/common.py
```diff
@@ -45,10 +45,10 @@ def course_delete(request, pk):
 
 
 def assignments(request):
-    public_list = models.Assignment.objects.filter(access='Public')
+    list = models.Assignment.objects.filter()
```
noisecapella: Using `list` as a variable name should be avoided since it's a builtin type. Can you rename to `assignments` or similar?

annagav:
Fixed!

noisecapella:
:+1:

PR #458, opened 2015-12-02
ab: homepage, create instructor account

annagav:


noisecapella:
:+1: 

This PR looks fine, but I noticed that anyone can become an instructor just by registering themselves, even on starcellbio.mit.edu. Is any protection against this planned? cc @pdpinch

pdpinch:
Yes, before we go to production.

> Is any protection against this planned?

ald32:
We haven't planned any protection to ensure that someone is actually an instructor. What are the potential downfalls to letting anyone make their own assignments?

PR #459, opened 2015-12-03
ab: fix file name addmultipledialog

annagav:


noisecapella:
Is this related to the rename in #442? I ran grep and found another reference to that file which may need changing:

```
(starcellbio)vagrant@vagrant-ubuntu-trusty-64:/vagrant$ grep addmutlpledialog . -R
./instructor/templates/instructor/preview.html:<script type="text/javascript" src="/static/assignments/scb_ex/gen/addmutlpledialog.soy.js?_=1448820534" charset="UTF-8"></script>
```

Can you update that as well? (unless it shouldn't be for some reason)

annagav:
I noticed that this file was misspelled after the rename in #442.

noisecapella:
:+1:

PR #460, opened 2015-12-03
Feature/footer funding link

annagav:
Alison asked to only leave the Funding link in the footer. Also updated url.


noisecapella:
:+1:

PR #477, opened 2015-12-07
ab compiler: add units to variables

### LINKED ISSUE #471: Assignment builder: compiler needs to read units for variables.
The compiler needs to read the units of all of the variables in the experiment setup and display them in the preview and regular modes.

annagav:
Fix #471

### DIFF
--- a/instructor/compiler.py
+++ b/instructor/compiler.py
@@ -353,12 +353,16 @@ def drugs(assignment):
 def concentrations(assignment):
     ret = {}
     for strain_protocol in assignment.strain_treatment.filter(enabled=True):
-        treatment = strain_protocol.treatment
-        concentration = treatment.drug.concentration
-        ret[str(concentration)] = {
-            'name': str(concentration),
-            'value': concentration
-        }
+        drug = strain_protocol.treatment.drug
+        concentration = drug.concentration
+        if not concentration is None:
+            ret[concentration] = {
+                'name': u"{concentration}{unit}".format(
+                    concentration=concentration,
+                    unit=drug.concentration_unit
+                ),
+                'value': concentration
+            }
     return ret
 
 
@@ -369,7 +373,7 @@ def experiment_temperatures(assignment):
             treatment = strain_protocol.treatment
             temperature = treatment.temperature
             ret[str(temperature.id)] = {
-                'name': str(temperature.degrees)
+                'name': u'{degrees}\u00b0C'.format(degrees=temperature.degrees)
             }
     return ret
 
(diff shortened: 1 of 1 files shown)


noisecapella:
Can you rebase please?

annagav:
rebased!

noisecapella:
It looks like it needs another rebase :)

annagav:
rebased again :)

instructor/compiler.py
```diff
@@ -353,10 +353,10 @@ def drugs(assignment):
...
     ret = {}
     for strain_protocol in assignment.strain_treatment.filter(enabled=True):
-        treatment = strain_protocol.treatment
-        concentration = treatment.drug.concentration
-        ret[str(concentration)] = {
-            'name': str(concentration),
+        drug = strain_protocol.treatment.drug
+        concentration = drug.concentration
+        ret[concentration] = {
+            'name': str(concentration) + drug.concentration_unit,
```
noisecapella: Instead of using string concatenation here can you use format strings? For example `"{concentration}{unit}".format(concentration=concentration, unit=drug.concentration_unit)`. Format strings are nicer for i18n and as a side benefit you don't have to use `str()` each time you want to concatenate a number with a string

annagav: Gives me an error: 'ascii' codec can't encode character u'\u03bc' in position 0: ordinal not in range(128)

noisecapella: Can you try a unicode format string instead?

annagav:
fixed!

html_app/model/Drugs.js
```diff
@@ -53,5 +53,5 @@ scb.Drug = function scb_Drug(data, context,parent) {
 	}, scb.utils.read_only_exception);
 
 	scb.Utils.accessor_toString(self.drug_id);
-	scb.Utils.accessor_toString(self.concentration_id);
+	//scb.Utils.accessor_toString(self.concentration_id);
```
noisecapella: This change looks new, was this intentional?

instructor/compiler.py
```diff
@@ -369,7 +373,7 @@ def experiment_temperatures(assignment):
             treatment = strain_protocol.treatment
             temperature = treatment.temperature
             ret[str(temperature.id)] = {
-                'name': str(temperature.degrees)
+                'name': str(temperature.degrees) + u'\u00b0C'
```
noisecapella: Can you use the format string here and below as well for `start_time` and `duration`?

noisecapella:
When clicking on 'Preview' on the home page I get this error:

screenshot from 2015-12-10 10 25 29

I think it was introduced in an earlier PR since it's a problem on `develop` too. I think the source is at `compiler.py:432`. Would you mind fixing that as part of this PR?

annagav:
Fixed!

noisecapella:
:+1:

PR #480, opened 2015-12-07
ab: order treatments by all vars

annagav:


noisecapella:
I noticed that collection time is unaffected by these changes FYI, but the PR itself looks good :+1:

Issue #462, opened 2015-12-03
Assignment builder preview: Update Add Samples dialog to new format


ald32:
Format the Add samples dialog window so that it matches the current format in SCB. This means that there should be 'Select all' and 'Clear all' buttons. Also the entire sample protocol name should appear including the strain, treatment, and all other relevant conditions.

annagav:
This is on scbdev. @ald32 , please make sure it looks (and works) as aspected.

ald32:
This looks great!

PR #486, opened 2015-12-08
ab: new addsamplesdialog

annagav:
#462

### DIFF
--- a/instructor/compiler.py
+++ b/instructor/compiler.py
@@ -11,6 +11,21 @@ def preview_as_json(assignment_id):
 
 
 
+def get_protocol_headers(assignment):
+    headers = ['Strain', 'Treatment']
+    # Optional headers
+    optional_vars = [('Concentration', 'has_concentration'),
+                     ('Start Time', 'has_start_time'),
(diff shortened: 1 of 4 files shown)


instructor/compiler.py
```diff
@@ -11,6 +11,23 @@ def preview_as_json(assignment_id):
...
+                     'Collection Time']
+    # assignment fields names
+    var_fields = ['has_concentration', 'has_start_time',
+                  'has_duration', 'has_temperature',
+                  'has_collection_time']
+
+    # Adding headers for enabled experimental variables
+    for index, field in enumerate(var_fields):
+        if getattr(assignment, field):
+            headers.append(optional_vars[index])
```
noisecapella: It would be cleaner to combine `optional_vars` and `var_fields` to be a list of tuples, for example `[('Concentration', 'has_concentration'), ('Start Time', 'has_start_time')`... This would allow you to iterate through it like

```
for header, field in pairs:
    ...
```

annagav:
Fixed!

noisecapella:
:+1:

PR #487, opened 2015-12-08
ab: create bands on setup change

### LINKED ISSUE #473: Assignment builder: band intensity page needs to update if there are changes to setup
If the instructor makes changes to the experiment setup, then those changes need to be reflected in the western blotting experimental technique - notably the band intensity page. Not sure if other pages need to be updated too.

annagav:
Fix #473


noisecapella:
I'm seeing a problem where if the band size textbox is empty and I update the value and save, the textbox clears the value and no bands are shown. I don't seem to have trouble updating existing band sizes however

noisecapella:
:+1:

PR #489, opened 2015-12-09
ab: filters for None variables

annagav:
Issue: After 'concentration', 'start time' and 'duration' were changed into IntegerField, empty/blank value were displayed as None.

### DIFF
--- a/instructor/templates/instructor/strain_protocols.html
+++ b/instructor/templates/instructor/strain_protocols.html
@@ -41,20 +41,20 @@
                     <div>{{ form.instance.treatment.drug.name }}</div>
                     {% if has_concentration %}
                         <div>
-                            {{ form.instance.treatment.drug.concentration }}
-                            {{ form.instance.treatment.drug.concentration_unit }}
+                            {{ form.instance.treatment.drug.concentration|default_if_none:'' }}
+                            {{ form.instance.treatment.drug.concentration_unit|default_if_none:'' }}
                         </div>
                     {% endif %}
                     {% if has_start_time %}
                         <div>
-                            {{ form.instance.treatment.drug.start_time }}
-                            {{ form.instance.treatment.drug.time_unit }}
+                            {{ form.instance.treatment.drug.start_time|default_if_none:'' }}
+                            {{ form.instance.treatment.drug.time_unit|default_if_none:'' }}
                         </div>
                     {% endif %}
                     {% if has_duration %}
                         <div>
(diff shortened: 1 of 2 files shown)


pdpinch:
What issue does this address?

annagav:
There was no issue for this. I just fixed it as soon as I noticed.

instructor/templates/instructor/strain_protocols.html
```diff
@@ -41,20 +41,20 @@
...
+                            {{ form.instance.treatment.drug.start_time|default_if_none:'' }}
+                            {{ form.instance.treatment.drug.time_unit|default_if_none:'' }}
                         </div>
                     {% endif %}
                     {% if has_duration %}
                         <div>
-                            {{ form.instance.treatment.drug.duration }}
-                            {{ form.instance.treatment.drug.time_unit }}
+                            {{ form.instance.treatment.drug.duration|default_if_none:'' }}
+                            {{ form.instance.treatment.drug.time_unit|default_if_none:'' }}
```
noisecapella, line 57: Do we need this for the `has_temperature` or `has_collection_time` blocks below?

noisecapella, line 57: Nevermind, the description explains that these three were changed to integers recently

instructor/templates/instructor/wb_band_intensity.html
```diff
@@ -34,11 +34,11 @@
                     <div class="scb_ab_s_col_width_sample">
                         <div class="block-ellipsis">
                         {{ form.instance.strain_protocol.strain.name }},
-                        {{ form.instance.strain_protocol.treatment.drug.name }}{% if variables.has_concentration %},
+                        {{ form.instance.strain_protocol.treatment.drug.name }}{% if variables.has_concentration and form.instance.strain_protocol.treatment.drug.concentration %},
```
noisecapella: The expression here checks if `drug.concentration` is falsy, not strictly that it's not none, so if you had a value `0` it would not be printed. Can you add `!= None` to the three expressions here to guard against this?

instructor/templates/instructor/strain_protocols.html
annagav, line 57: This does not apply to 'temperature' and 'collection_time' because these variables, if present, cannot be empty.

annagav:
Fixed!

noisecapella:
:+1:

PR #490, opened 2015-12-10
ab: access only owner's assignments and courses

annagav:
Filter assignments and courses by owner. 
Also removed two unused views.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -14,38 +14,11 @@ from django.core.exceptions import PermissionDenied
 import json
 import re
 from instructor.compiler import get_protocol_headers
-@login_required
-def courses(request):
-    message = ''
-    CourseFormSet = modelformset_factory(models.Course, extra=1, fields=['name', 'code'], can_delete=True)
-    if request.method == "POST":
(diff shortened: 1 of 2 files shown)


noisecapella:
When I step through the application as a new user I don't see other user's courses or assignments, which is fine, but there are still endpoints which aren't checking ownership properly. For example anyone can go to `/ab/assignments/delete/34` and delete assignment 34 without owning it. `publish_assignment` is also unrestricted.

noisecapella:
By the way if you want to get this PR merged and deal with the other security issues later, that's fine with me

instructor/common.py
```diff
@@ -131,7 +104,7 @@ def course_setup(request):
                 course.save()
 
     # if there is at least one course
-    all_courses = models.Course.objects.all()
+    all_courses = models.Course.objects.filter(owner__username=request.user)
```
noisecapella: It seems like this should compare against `request.user.username`

annagav:
Let's do the rest in another PR :)

noisecapella:
:+1:

PR #495, opened 2015-12-11
ab: order samples in addsamples dialog

annagav:
The order of samples in 'add samples' dialog should be the same as listed in the protocols page.


noisecapella:
:+1:

Issue #449, opened 2015-11-30
Western blot band size page: Minor edit to 'WCL'


ald32:
The heading should read 'WCL' not 'WLC'

PR #497, opened 2015-12-11
ab: WLC to WCL

annagav:
#449


noisecapella:
No idea what WCL is but seems simple enough :+1:

PR #500, opened 2015-12-11
ab compiler: wb format table

annagav:


noisecapella:
For the record I currently get: `s1, t1, 3μg/μL, 3min, [object Object]`
before it was `s1, t1, [object Object]`

But it's fine for this PR :+1:

PR #513, opened 2015-12-16
ab: error if course code exists

### LINKED ISSUE #510: Course setup: Display an error message when the course code is not unique
(I'm not sure if there is already an issue for this- I apologize if this is a duplicate)

If an instructor does not use a unique course code, then the following message should appear:

The selected course code cannot be used as it is not unique within StarCellBio's course code list. Please use another course code.

'OK'

When this error message is triggered, the course name that the instructor typed in should be saved, but the course code should be deleted.

annagav:
Show formset errors. Fix #510.


noisecapella:
:+1:

PR #515, opened 2015-12-16
ab: delete, publish, check owner

annagav:
Check owner when deleting or publishing an assignment.


noisecapella:
This looks good but there's still more work to be done to secure authorization for the application, for example a user can still edit an assignment without owning it. :+1:

PR #518, opened 2015-12-16
ab: antibody placeholders

### LINKED ISSUE #508: Antibody page: put placeholder text in the text box
I think we should provide a suggestion to the instructor for what the primary and secondary antibody names should be. Can you put the following text in the appropriate text boxes - in a gray color so the instructor knows that they are only hints and not actual defined names?

Primary antibody: 
goat anti-protein

Secondary antibody:
rabbit anti-goat

annagav:
Fix #508

### DIFF
--- a/instructor/ui/instructor.js
+++ b/instructor/ui/instructor.js
@@ -180,6 +180,15 @@ $(function(){
         }
     });
     /**
+     * Western Blotting: Antibodies
+     */
+    if($("#id_form-0-primary").length){
+        if($(".scb_ab_s_form_input_list").length < 2){
+            $("#id_form-0-primary").attr('placeholder', 'goat anti-protein');
+            $("#id_form-0-secondary").attr('placeholder', 'rabbit anti-goat');
+        }
+    }
+    /**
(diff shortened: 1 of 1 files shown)


instructor/ui/instructor.js
```diff
@@ -180,6 +180,15 @@ $(function(){
         }
     });
     /**
+     * Western Blotting: Antibodies
+     */
+    if($("#id_form-0-primary").length){
```
noisecapella, line 185: Are these CSS classes specific to the western blotting antibodies page? If not can you add one to prevent this code from executing elsewhere?

noisecapella:
Just the one question there, functionality looks good otherwise. It looks like Django Forms provide a way to set the placeholder but it doesn't look easy to do if you're starting with a modelformset

instructor/ui/instructor.js
annagav, line 185: Yes, this input id for primary antibody is unique to this page.

noisecapella:
:+1:

PR #520, opened 2015-12-17
ab: wb re-probing bug

annagav:
Assignments Builder Preview: re-probing was showing the same marks as in the first blot.


pdpinch:
:+1:

PR #523, opened 2015-12-21
ab: enable pages one by one

### LINKED ISSUE #448: Assignment builder: prevent navigation to pages that the instructor hasn't yet reached
The instructor should only be able to navigate back to pages that s/he has already completed, and should not access pages that have not been reached via navigation through using the 'Save & Continue' buttons. This means that the applicable lines should be grayed out in the left navigation menu.

annagav:
Fix #448

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -74,23 +74,24 @@ def assignment_setup(request):
                 break
 
     all_assignments = models.Assignment.objects.filter(course__owner=request.user)
-
-    return render_to_response('instructor/assignment_setup.html',
-                              {
-                                  'assignments': all_assignments,
-                                  'access': 'private',
-                                  'error': error,
-                                  'assignment_name': assignment_name,
-                                  'based_on': request.session['based_on'],
-                                  'new': request.session['new'],
-                                  'section_name': 'Assignment',
-                                  'page_name': 'assignment'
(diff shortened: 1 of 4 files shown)


instructor/common.py
```diff
@@ -680,6 +737,9 @@ def western_blot_band_size(request):
             update_wb_bands(assignment, antibodies=antibodies)
 
             if 'continue' in request.POST:
+                if assignment.last_enabled_page <= page_number:
+                    assignment.last_enabled_page = page_number+1
+                    assignment.save()
```
pdpinch, line 786: You have this code block in a lot of places. You may want to refactor it into a helper of some kind.

pdpinch:
Just to make sure I understand, you're now storing a "last_enabled_page" on the instructor assignment. It gets incremented for each page, and then you check it to decide which pages to enable?

The number always goes up, so someone who returns to edit a problem after filling out every form will be able to jump in wherever they like?

One concern is that if you need to insert a new form in the middle you'll have to renumber all the subsequent forms -- that's a lot of  tweaking in common.py and the template. But I don't think we'll be doing that anytime soon, so it's probably not worth coming up with a more sophisticated approach (at least not yet). 

It's your call on the refactoring. 

:+1:

PR #524, opened 2015-12-21
ab: publish assignment

### LINKED ISSUE #501: Publish assignment so student can view
- [x] compile assignment, similar to preview
- [x] save resulting json to student table

annagav:
Fix #501

### DIFF
--- a/StarCellBio/views.py
+++ b/StarCellBio/views.py
@@ -273,16 +273,10 @@ def get_student_courses(request, **kwargs):
         else:
             all = []
             # sample courses
-            #for a in Assignment.objects.filter(courseID=Course.objects.filter(code='SCB_SampleExercises')):
-            for a in Assignment.objects.all():
+            for a in Assignment.objects.filter(courseID=Course.objects.filter(code='SCB_SampleExercises')):
                 dictionary = ast.literal_eval(a.data)
                 all.append(dictionary)
 
-            # for a in Assignment.objects.all():
-            #     dictionary = ast.literal_eval(a.data)
-            #     if (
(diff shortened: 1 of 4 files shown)


pdpinch:
I'm having problems with publishing some experiments, which I think may be bugs in the compiler. 

The first problem I had was with an experiment with only one strain. When I got to the add strains portion of the setup, I didn't get a pop-up window at all. No console messages. 

In trying to reproduce that error, I'm now getting a blank screen after I sign into my course as a student. Looking at the console I see a message:

```

              <span style="color: #000000; font-size: small;">This
              feature is currently unavailable in
              StarCellBio.</span>

              <span style="color: #000000; font-size: small;"><span style="line-height: 21px;"><br></span></span>

              <span style="font-size: small;"><br></span>
```

But it doesn't get to the user. 

I'm not sure if it's related, but I can't get the json stored in the backend.assignments table to validate. One thing that looks especially suspicious is that the values for drug_id aren't quoted.

annagav:
Have you defined the treatments and the western blot technique?

pdpinch:
I have defined the treatments and the western blot technique. 

It actually seems to happen when I have more than one assignment in the same course. One assignment is OK, but two assignments blows up. 

Otherwise, this appears to be working OK.

annagav:
Rebased and fixed!

annagav:
@pdpinch, Could you please take a look at it again, so that I can rebase PR #532?

pdpinch:
OK. Doing it now.

pdpinch:
Ok, this is working for me now. I published two assignments, each with only one strain. I was able to do western blots for both of them.

:+1:

PR #528, opened 2015-12-22
ab: view published assignment

### LINKED ISSUE #484: Create 'View' functionality for published assignments
When the instructor publishes an assignment, 'Edit' should change to 'View' in the Actions table on the dashboard. The instructor should be able to view the assignment builder in a read-only way when s/he clicks on 'View' in the actions column.

annagav:
Fix #484

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -78,6 +78,7 @@ def assignment_setup(request):
     return render_to_response('instructor/assignment_setup.html',
                               {
                                   'assignments': all_assignments,
+                                  'access': 'private',
                                   'error': error,
                                   'assignment_name': assignment_name,
                                   'based_on': request.session['based_on'],
@@ -121,6 +122,7 @@ def course_setup(request):
     return render_to_response('instructor/course_modify.html',
                               {
                                   'courses': all_courses,
+                                  'access': 'private',
                                   'formset': formset,
                                   'course_selected': course_selected,
                                   'new': request.session['new'],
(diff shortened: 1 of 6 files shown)


instructor/common.py
```diff
@@ -672,20 +707,22 @@ def western_blot_band_size(request):
         extra=0,
         exclude=exclude_fields
     )
-    if request.method == "POST":
-        weights_by_type = ['wc_weight', 'nuc_weight', 'cyto_weight']
```
pdpinch, line 685: We don't need this anymore?

instructor/templates/instructor/assignments.html
```diff
@@ -46,7 +46,11 @@ <h1 class="scb_s_dashboard_sidebar_title">Your Assignments: </h1>
                         </td>
                         <td class='scb_s_dashboard_table_element scb_s_dashboard_table_border'>
                             <a class='scb_ab_s_dashboard_action'
-                               href="{% url "common_assignments_edit" assignment_pk=a.id %}">Edit</a>
+                               href="{% url "common_assignments_edit" assignment_pk=a.id %}">
```
pdpinch, line 49: don't we want to check the URL here to be the view URL instead of the edit URL?

instructor/common.py
annagav, line 685: Yes, old residual :)

instructor/templates/instructor/assignments.html
annagav, line 49: There is only one view for both.

pdpinch, line 49: When the button changes to `view` I was expecting that clicking it would take me to the student view, not back to editing.

annagav, line 49: Were you able to edit something?

annagav:
Rebased

instructor/templates/instructor/assignments.html
annagav, line 49: The 'view' mode for the instructor is the same as the 'edit' mode except that you can't change any values.

pdpinch:
Ok. I'm not crazy about this UX, but this is OK to merge. :shipit:

PR #529, opened 2015-12-22
ab: empty vars bug

### LINKED ISSUE #461: Assignment builder preview mode: remove [object object] on WB pages
Remove [object object] that appears after the sample name in the sample window on the western blotting experimental technique pages. This appears on the lysate prep pages and in the samples window on the Prepare/Load pages. Can't remember if this also appears once you run gel & transfer.

annagav:
Fix #461

### DIFF
--- a/html_app/model/CellTreatment.js
+++ b/html_app/model/CellTreatment.js
@@ -93,6 +93,7 @@ scb.CellTreatment = function scb_CellTreatment(data, context, parent) {
                 new_text = new_text.replace( regexp , resolve(value));
             });
             display_text = new_text == orig_template ? display_text : new_text;
+            display_text = display_text.replace(/(, *)+/g, ", ");
         }
         return display_text;
     };
--- a/instructor/compiler.py
+++ b/instructor/compiler.py
@@ -120,7 +120,6 @@ def csv_custom(data):
 
 
 def format_table(assignment):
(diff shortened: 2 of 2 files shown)


pdpinch:
Not sure how to test this. I'll come back to it later.

annagav:
The bug was happening when some of variables, like start time, concentration, or duration, were left empty in the assignment builder, and later in the preview of the technique they appeared as [object object].

html_app/model/CellTreatment.js
```diff
@@ -93,6 +93,7 @@ scb.CellTreatment = function scb_CellTreatment(data, context, parent) {
                 new_text = new_text.replace( regexp , resolve(value));
             });
             display_text = new_text == orig_template ? display_text : new_text;
+            display_text = display_text.replace(/(, *)+/g, ", ");
```
pdpinch, line 96: This seems a little odd to remove the extra commas after the fact. Where does the display_text get constructed? Is there a way to omit the empty fields at that point?

Not a big deal. If it's too painful to change, I'm :+1:

annagav:
display_text is being constructed right above this. It uses a big chunk of assignment structure. It would involve a lot of change and refactoring to change the way display_text is computed. I don't think it is needed at this point.

PR #530, opened 2015-12-23
ab: show message before publishing

### LINKED ISSUE #481: Dashboard: when publishing an assignment, show a confirmation message
As an instructor, I need to know that when I publish an assignment, I will no longer be able to edit it.

annagav:
Fix #481


pdpinch:
:+1: on the message, but I can still edit after I publish & the state changes to public. Is that what I should expect?

Also, is there any way to return to private?

annagav:
There is no way to return to private.

annagav:
The actual 'view' functionality is in different PR.

pdpinch:
I see that now. This is good to go now.

PR #532, opened 2015-12-28
ab: add input box for duration time

### LINKED ISSUE #446: Define variables: start time needs its own units
The start time variable needs its own units and can't share a units dropdown with duration.

annagav:
Fix #446

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -442,16 +442,17 @@ def assignments_edit_treatments(request):
     else:
         input_headers.extend(['Concen.', 'Concen. Units'])
 
-    if not (has_duration or has_start_time):
-        drug_formset_exclude.extend(['duration', 'start_time', 'time_unit'])
-    elif not has_duration:
+    if has_duration and has_start_time:
+        input_headers.extend(['Start Time', 'Time Units', 'Duration', 'Time Units'])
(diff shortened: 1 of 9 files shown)


instructor/common.py
```diff
@@ -449,7 +450,7 @@ def assignments_edit_treatments(request):
         'g/L',
         'nM',
         u'\u03BCM', #uM
-        'mm',
+        'mM',
```
pdpinch, line 538: is this the abbreviation for millimeters??

pdpinch, line 538: Nevermind -- it's a concentration not a distance. It looks like it's millimolar.

pdpinch:
I'm pretty sure this is working. I can't test it on a published course because I'm waiting on #524 and #528 so I can't confirm if the changes to compiler.py are working. 

I can test again after the others are merged and this is rebased.

annagav:
Rebased!

pdpinch:
This is working both in preview and in student view. 

One thing I just noticed though: it would be nice if there were a space between the value & the units:

screen shot 2015-12-30 at 2 17 04 pm

You can either fix that, or merge this and open an issue for the spacing.

Otherwise :+1:

PR #535, opened 2016-01-04
ab: preview and publish check assignment

### LINKED ISSUE #506: Prevent instructors from previewing their assignment before it is completely defined
An instructor should not be able to preview their assignment if they try to 'Preview' their assignment before the experiment setup and one experimental technique have been defined. A dialog window should appear that says:
You must define your experimental setup and the variables for at least one experimental technique before you can preview the assignment.
'OK'

annagav:
Fix #506

### DIFF
--- a/StarCellBio/urls.py
+++ b/StarCellBio/urls.py
@@ -69,7 +69,7 @@ urlpatterns += patterns(
     url(r'^ab/assignments/$', instructor_common.assignments, name="common_assignments"),
     url(r'^ab/assignments/delete/(?P<assignment_pk>.+)$', instructor_common.assignment_delete,
         name="common_assignment_delete"),
-    url(r'^ab/assignments/publish/(?P<assignment_pk>[^/]+)$', instructor_common.publish_assignment,
+    url(r'^ab/assignments/publish/$', instructor_common.publish_assignment,
         name="common_publish_assignment"),
 
     # Creating assignment
@@ -99,8 +99,11 @@ urlpatterns += patterns(
         instructor_common.assignments_variables, name="common_assignments_variables"),
 
     # Preview
-    url(r'^ab/assignments/preview/(?P<assignment>[^/]+)$', instructor_common.preview,
+    url(r'^ab/assignments/preview/(?P<assignment_pk>[^/]+)$', instructor_common.preview,
         name="common_preview"),
+    # Check if assignment is complete
+    url(r'^ab/assignments/assignment_complete/$', instructor_common.assignment_complete,
+        name="common_assignment_complete"),
 
     # Select Technique
     url(r'^ab/assignments/select_technique/$', instructor_common.select_technique,
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -15,18 +15,26 @@ import backend.models
 import json
 import re
 from instructor.compiler import get_protocol_headers
-
+from django.http import HttpResponse, HttpResponseBadRequest
 
 @login_required
 def assignments(request):
     assignments = models.Assignment.objects.filter(course__owner=request.user)
+    return render_to_response(
+        'instructor/assignments.html',
+        {
+            'assignments': assignments
+        },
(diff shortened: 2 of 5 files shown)


instructor/common.py
```diff
@@ -18,30 +18,41 @@
...
+                courseID=course,
+                assignmentID=assignment.id,
+                assignmentName=assignment.name,
+                data=assignment_json,
+                ownerID=assignment.course.owner,
+                access=assignment.access
+            )
+            backend_assignment.save()
+        return redirect('common_assignments')
+    else:
```
justinabrahms: I'd recommend inverting this else, so you lose a level of indentation.

``` python
if not assignment_ready:
  return redirect('common_assignments, error="...")
if request.user....
```

instructor/common.py
```diff
@@ -18,30 +18,41 @@
 
 
 @login_required
-def assignments(request):
+def assignments(request, error=''):
     assignments = models.Assignment.objects.filter(course__owner=request.user)
+    return render_to_response(
+        'instructor/assignments.html',
+        {
+            'error': json.dumps(error),
```
justinabrahms: You have `dumps` here, which takes a python object and turns it to a string, but the urlresolver will only ever give you strings.

Also, accepting json as a url fragment instead of as a post variable or query param seems quite odd. Also, `/` is valid for json contents. Perhaps I'm missing some important context here.

noisecapella: Just adding this since I made this suggestion in the first place. Since the error is being used as a variable in Javascript wouldn't it make sense to JSON encode rather than use the string escaping in Django templates?

StarCellBio/urls.py
```diff
@@ -67,9 +67,10 @@
 
     # Dashboard
     url(r'^ab/assignments/$', instructor_common.assignments, name="common_assignments"),
+    url(r'^ab/assignments/(?P<error>[^/]+)$', instructor_common.assignments, name="common_assignments"),
```
noisecapella: What is this for? It points to the same view as the previous line and I don't see any reference to `error` in the view

annagav: Not needed anymore :)

instructor/common.py
```diff
@@ -1050,7 +1061,27 @@ def facs_histograms_edit(request, assignment, sample_prep, sp):
...
+            {
+                'assignment': a,
+                'assignment_json': compiler.preview_as_json(a.id)
+            },
+            context_instance=RequestContext(request))
+    else:
+        error = "You must define your experimental setup and the variables " \
+                "for at least one experimental technique before you can preview " \
+                "the assignment."
+        return redirect('common_assignments', error=error)
```
noisecapella: I get this error when I get to this case:

```
Reverse for 'common_assignments' with arguments '()' and keyword arguments
 '{'error': 'You must define your experimental setup and the variables for at least
 one experimental technique before you can preview the assignment.'}' not found. 
1 pattern(s) tried: ['ab/assignments/$']
```

instructor/common.py
```diff
@@ -15,18 +15,26 @@
...
-                              {'assignments': assignments},
-                              context_instance=RequestContext(request))
 @login_required
-def publish_assignment(request, assignment_pk):
+def publish_assignment(request):
+    assignment_pk = request.POST.get('pk')
     assignment = get_object_or_404(models.Assignment, pk=assignment_pk)
+    assignment_ready = is_assignment_complete(assignment)
+    if not assignment_ready:
+        return HttpResponse("Cannot publish unfinished assignment.")
```
noisecapella: Can you return an error status code in this case, a 400 would work here. Then you can change your jQuery below to something like

```
$.ajax(...).then(function() {
  window.location = ...
}).fail(function(message) {
  show_alert(message);
});
```

In general it's important to use status codes in representing errors so that other libraries can know what is an error and what isn't without parsing your message

annagav:
Fixed!

noisecapella:
:+1:

Issue #482, opened 2015-12-08
Dashboard, change values of Access column to private/published


pdpinch:
As an instructor, it would be clearer if the "publish" action resulted in the access status changing to "published" instead of "public"

PR #550, opened 2016-01-08
ab: warn when deleting assignment

### LINKED ISSUE #537: Warn instructor if they delete an assignment on the dashboard
I think that an instructor should be warned that if they delete an assignment, it will be removed permanently. If the assignment is private, then the confirmation dialog box should say:
Your assignment will be removed permanently from your dashboard. Would you like to continue?

If the assignment is public, then the confirmation dialog box should say:
Your assignment will be removed permanently from your dashboard and students will no longer be able to access this assignment within this course and their work will be deleted. Would you like to continue?

Yes/No

annagav:
Fix #537, #482

### DIFF
--- a/instructor/migrations/0015_published_assignment.py
+++ b/instructor/migrations/0015_published_assignment.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('instructor', '0014_drug_duration_unit'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='assignment',
+            name='access',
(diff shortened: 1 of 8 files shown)


instructor/migrations/0015_published_assignment.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('instructor', '0014_drug_duration_unit'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='assignment',
+            name='access',
+            field=models.CharField(default=b'private', max_length=50, choices=[(b'published', b'Published'), (b'private', b'Private'), (b'archived', b'Archived')]),
```
noisecapella, line 17: This migration won't change existing fields that already have the value `public` in them. You will need a data migration for that: https://docs.djangoproject.com/en/1.9/topics/migrations/#data-migrations

annagav:
Fixed!

noisecapella:
:+1:

PR #552, opened 2016-01-12
ab: adding extra assignments to existing course

### LINKED ISSUE #542: ab: adding new assignments to a course with existing student accounts doesn't work
I added a second assignment to an already existing course and then tried to view the assignment in a student account that I had made to view the first assignment in this course. I just got a blank screen and the page wouldn't load. However, I was able to make a new student account and see both assignments in the course.

annagav:
Fix #542

### DIFF
--- a/StarCellBio/views.py
+++ b/StarCellBio/views.py
@@ -229,23 +229,7 @@ def get_student_courses(request, **kwargs):
                         a = Assignment.objects.filter(assignmentID=x)
                         a = a[0]
                         original_assignment_data = a.data
-                        assignment_data = ast.literal_eval(original_assignment_data)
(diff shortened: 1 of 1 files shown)


noisecapella:
How should I reproduce the bug? Does the course or assignment have to have a special name?

annagav:
To reproduce the bug:
1. As an instructor: publish an assignment (with some course code)
2. As a student login/create account with this course code. The student should see the assignment(s) for this course.
3. As an instructor: create a new assignment with the course code above or change the course for an existing assignment. Then publish this assignment. Now there is one new assignment for that course.
4. As a student: login to that course again or refresh the page.

noisecapella:
Ok, looks good :+1:

PR #559, opened 2016-01-15
ab: assignment check owner

annagav:


instructor/common.py
```diff
@@ -195,6 +195,18 @@ def assignments_edit(request, assignment_pk):
 
     return redirect('common_assignment_modify')
 
+
+def check_assignment_owner(func):
+    def check_owner(request):
```
noisecapella: The decorator looks good overall. You should also use `functools.wraps` to ensure that you retain the name and docstring of the function being decorated: http://stackoverflow.com/questions/308999/what-does-functools-wraps-do

instructor/common.py
```diff
@@ -195,6 +195,18 @@ def assignments_edit(request, assignment_pk):
...
+        assignment_id = request.session['assignment_id']
+        assignment = models.Assignment.objects.get(id=assignment_id)
+        user = get_object_or_404(User, username=request.user)
```
noisecapella, line 206: This should be `request.user.username` I think

instructor/common.py
noisecapella: Also, you should replace `request` with `*args, **kwargs` and pass it to `func` below instead of `func(request)`

instructor/common.py
annagav, line 206: request.user seems to have the username.

instructor/common.py
annagav: Fixed

instructor/common.py
noisecapella, line 206: You're right, I guess Django is converting `User` to a `str` before doing the query, and since `str(user)` is the username this works.

noisecapella:
:+1: Nice use of a decorator!

annagav:
Thank you!

Issue #445, opened 2015-11-30
On the protocols page, the temperature units should not be in bold


ald32:
The degrees C text should not be in bold. Is this unit text in the same font and size as the rest of the text in the protocols table?

PR #564, opened 2016-01-21
ab: intensity page format

### LINKED ISSUE #452: Band intensity page: formatting and edits for the table
The table should be formatted like other tables where the headings are in bold, etc. so that it is clear that they are headings. 

It is hard to see the numbers next to each sample. Can you indent the text slightly (in both the first and second rows) so that it is easy to see each sample number (and tell that the sample name goes onto two lines)? 

The lysate type, size (kDa), and relative band intensity appear to be vertically bottom aligned within each row. Can you vertically center these items in each row?

In the heading for the Size column, please add a space between 'Size' and '(kDa)'. It should read: 'Size (kDa)'

annagav:
Fix #452, #445

### DIFF
--- a/instructor/templates/instructor/strain_protocols.html
+++ b/instructor/templates/instructor/strain_protocols.html
@@ -27,7 +27,7 @@
         <form method="post">{% csrf_token %}
             {{ formset.management_form }}
 
-            <div class="scb_ab_s_row scb_ab_s_col_width">
+            <div class="scb_ab_s_row scb_ab_s_col_width scb_ab_s_table_header_text">
                 <div class="scb_ab_s_enable_sample">Select</div>
             {% for header in headers %}
                 <div>{{ header }}</div>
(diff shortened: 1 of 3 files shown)


noisecapella:
Is this part done? I'm not sure what to look for exactly. Everything else looks good to me

> It is hard to see the numbers next to each sample. Can you indent the text slightly (in both the first and second rows) so that it is easy to see each sample number (and tell that the sample name goes onto two lines)?
> 
> The lysate type, size (kDa), and relative band intensity appear to be vertically bottom aligned within each row. Can you vertically center these items in each row?

annagav:
The numbers are separate from the text. So I think that is done. 
As for the vertical alignment, I think having it aligned with the first row makes more clear. It didn't look good when aligned in the center, since the second row is not full and some rows don't have second line.

noisecapella:
Sounds good to me :+1:

noisecapella:
FYI I think you need to put the Fix statements on separate lines. It doesn't look like #445 was closed by this merge

PR #566, opened 2016-01-21
ab: vars positive integers

### LINKED ISSUE #549: ab: prevent negative numbers in experiment setup
I accidentally used the up and down arrows to enter a duration and was able to select a negative number and then continue to the next page. I thought that we were checking to make sure that all the numbers were positive (for concentration, duration, start time, and collection time). Temperature can be a negative number so that should be allowed.

annagav:
Fix #549


noisecapella:
:+1:

PR #568, opened 2016-01-28
ab: facs sample prep

### LINKED ISSUE #414: Order of techniques in left hand navigation should match order in select techniques form

### LINKED ISSUE #569: AB: FACS Sample Prep page
Define combinations of cell treatments, analysis and conditions. Design document: Problem Set Builder/User Interface Design Documents - Jamie Waters/SCB_ProblemSetBuilder_09_FlowCyt.pdf

annagav:
LowCytometry Sample Prep Page. Define cell treatment, analysis and condition for each sample.
Fix #569 
Fix #414

### DIFF
--- a/instructor/models.py
+++ b/instructor/models.py
@@ -41,7 +41,7 @@ class Course(models.Model):
 class Assignment(models.Model):
     course = models.ForeignKey(Course, related_name='assignments')
     assignment_id = models.CharField(max_length=50, unique=True)
-    last_enabled_page = models.IntegerField(max_length=5, default=2)
+    last_enabled_page = models.IntegerField(max_length=5, default=1)
     name = models.CharField(max_length=50)
     access = models.CharField(max_length=50, choices=ACCESS, default='private')
     basedOn = models.ForeignKey("Assignment", null=True, blank=True)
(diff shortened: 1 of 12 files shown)


pdpinch:
Hi Anna. I'm having a problem with this one. I can check Flow Cytometry in the Select Techniques form, but I can't access the Sample Prep form. Maybe we can look at it together tomorrow.

pdpinch:
Anna and I spoke. I was blocked because I didn't know about https://github.com/starteam/starcellbio_html/issues/580

pdpinch:
Is it OK that both fixed cells and live cells are unchecked, or do we need to validate that one of them is chosen? 

(If one must be chosen, traditional UX would be to use radio buttons instead of checkboxes)

instructor/models.py
```diff
@@ -41,7 +41,7 @@ def __unicode__(self):
 class Assignment(models.Model):
     course = models.ForeignKey(Course, related_name='assignments')
     assignment_id = models.CharField(max_length=50, unique=True)
-    last_enabled_page = models.IntegerField(max_length=5, default=2)
+    last_enabled_page = models.IntegerField(max_length=5, default=1)
```
pdpinch, line 44: Why this change?

annagav, line 44: Page numbers start with 0 now.

annagav:
It is ok not to specify the cell treatment ( at least for now). We probably don't want it as a radio button since both options can be selected.

instructor/models.py
```diff
@@ -168,17 +168,17 @@ class MicroscopyImages(models.Model):
 
 FACS_FIXED = 'Fixed'
 
-FACS_KINDS = (( 'Dye', 'Dye/Stain' ), ('Anti', 'Antibody-labeling'))
+FACS_KINDS = (('', 'Select Analysis'), ('Anti', 'Antibody-labeling'), ('Dye', 'Dye/Stain'))
 
 FACS_DYE = 'Dye'
```
pdpinch: Nit: If `FACS_FIXED` and `FACS_DYE` are meant to be keys, it might be clearer to use them in FACS_KINDS

pdpinch:
:+1:

instructor/models.py
annagav: I don't see any use for FACS_FIXED and FACS_DYE. I will remove them.

PR #575, opened 2016-02-01
Feature/ab facs analyze

### LINKED ISSUE #572: AB: FACS Analyze page - select histogram form
from https://www.dropbox.com/s/5dy257x6e4kc5sb/SCB_ProblemSetBuilder_09_FlowCyt.pdf?dl=0: 

Just this form: 

image

annagav:
Fix #572


instructor/common.py
```diff
@@ -1097,6 +1100,88 @@ def facs_sample_prep(request):
         context_instance=RequestContext(request))
 
 
+def create_facs_histograms(assignment, facs_sample):
```
giocalitri, line 1125: can you please add a docstring describing what this function does (with more details than just the function name)?

instructor/common.py
```diff
@@ -1097,6 +1100,88 @@ def facs_sample_prep(request):
...
+            sample_prep=facs_sample,
+            strain_protocol=strain_protocol,
+        )
+
+
+@check_assignment_owner
+@login_required
+def facs_analyze(request):
+    page_number = page_order.index('facs_analyze')
+    pk = request.session['assignment_id']
```
giocalitri, line 1153: question: can this potentially raise a `KeyError`?

instructor/common.py
```diff
@@ -1097,6 +1100,88 @@ def facs_sample_prep(request):
...
+    for strain_protocol in strain_protocols:
+        histogram, created = models.FlowCytometryHistogram.objects.get_or_create(
+            sample_prep=facs_sample,
+            strain_protocol=strain_protocol,
+        )
+
+
+@check_assignment_owner
+@login_required
+def facs_analyze(request):
```
giocalitri, line 1146: missing docstring

instructor/common.py
```diff
@@ -1097,6 +1100,88 @@ def facs_sample_prep(request):
...
+            'strain_protocol__treatment__collection_time__time'
+        ))
+    grouped_forms = []
+    samples = models.FlowCytometrySamplePrep.objects.filter(assignment=assignment)
+    for sample in samples:
+
+        form_list = [form for form in formset
+                     if form.instance.sample_prep.analysis == sample.analysis and
+                     form.instance.sample_prep.condition == sample.condition
+        ]
```
giocalitri, line 1194: nit: indent this better

instructor/common.py
```diff
@@ -1097,6 +1100,88 @@ def facs_sample_prep(request):
...
+            entries = formset.save(commit=False)
+            for entry in entries:
+                entry.save()
+            if 'continue' in request.POST:
+                if assignment.last_enabled_page <= page_number:
+                    assignment.last_enabled_page = page_number+1
+                    assignment.save()
+                return redirect('common_assignments')
+    elif request.method == "POST" and 'continue' in request.POST:
+        return redirect('common_assignments')
```
giocalitri, line 1174: Nit: I would be this condition as the first of the 3, considering it has the simplest code (but feel free to leave as it is)

instructor/common.py
```diff
@@ -1097,6 +1100,88 @@ def facs_sample_prep(request):
...
+    for sample in samples:
+
+        form_list = [form for form in formset
+                     if form.instance.sample_prep.analysis == sample.analysis and
+                     form.instance.sample_prep.condition == sample.condition
+        ]
+        if sample.fixed:
+            grouped_forms.append(('fixed', sample.analysis, sample.condition, form_list))
+        if sample.live:
+            grouped_forms.append(('live', sample.analysis, sample.condition, form_list))
```
giocalitri, line 1198: question: this piece of code might be heavy on memory (it has an N^2 complexity): do you have a sense of how many items `formset` and `samples` may have?

instructor/migrations/0021_fixed_live_histogram.py
```diff
@@ -0,0 +1,29 @@
...
+    dependencies = [
+        ('instructor', '0020_facs_sample_prep'),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name='flowcytometryhistogram',
+            old_name='data',
+            new_name='fixed_data',
+        ),
```
giocalitri, line 18: question: is this part of the migration (`migrations.RenameField`) correct or it is just the result of django interpreting in the wrong way the new code?

instructor/templates/instructor/assignment_sidebar.html
```diff
@@ -147,6 +147,16 @@
                 Sample Prep
             </a>
         </div>
+        <div role="link" class='scb_ab_s_sidebar_page_link
+            {% if page_name == "facs_analyze" %}scb_ab_s_sidebar_page_name_selected{% endif %}'>
+            <a class="scb_s_assignment_sidebar_link {% if last_page_number < 12 %} disabled {% endif %}"
```
giocalitri, line 152: question (because I really need some background): why "12"?

giocalitri:
Few minor changes and few questions asked.

Please add another commit to be later squashed to address the comments.

instructor/templates/instructor/assignment_sidebar.html
annagav, line 152: because it is the 13th page in the sidebar menu.

instructor/migrations/0021_fixed_live_histogram.py
annagav, line 18: I think rename field or delete and create would be equivalent here, since there is no data in the table.

instructor/common.py
annagav, line 1153: yes, it can. I am adding a decorator for that.

instructor/common.py
annagav, line 1194: I tried different ways to indent this but it does not look much different. Maybe you have a suggestion.

instructor/common.py
annagav, line 1198: however many samples and cell treatments the instructor will define. It is an arbitrary number.

giocalitri, line 1198: well, then at least use an iterator for the `samples` queryset. see https://docs.djangoproject.com/en/1.9/ref/models/querysets/#iterator

instructor/migrations/0021_fixed_live_histogram.py
giocalitri, line 18: ok

instructor/templates/instructor/assignment_sidebar.html
giocalitri, line 152: and what does it mean?

instructor/common.py
giocalitri, line 1194: something like this probably

```
form_list = [
    form for form in formset
    if form.instance.sample_prep.analysis == sample.analysis and
        form.instance.sample_prep.condition == sample.condition
]
```

giocalitri, line 1194: you already changed in this way: good job

instructor/templates/instructor/assignment_sidebar.html
annagav, line 152: last_page_number contains the value of the last enabled page that the can see/click on. So there is a check to disable the link to the page.

giocalitri, line 152: Sorry, I still do not understand why it should be 12: is it a requirement?

annagav, line 152: I numbered the pages based on their order, why would it be any other number than 12?

giocalitri, line 152: Anna, here you are assuming I have any knowledge of any kind on the number of pages for this project: this assumption is wrong.
If I need to acquire such knowledge before reviewing this PR, please point me to the right issue #, documentation or code.
cc @pdpinch

annagav, line 152: Giovanni, I think I have answered your question, why this particular number was chosen, without any assumptions on background knowledge. If your question was regarding the functionality of the code segment itself, please see this PR  https://github.com/starteam/starcellbio_html/pull/523

instructor/common.py
annagav, line 1125: added below

instructor/common.py
annagav, line 1198: good to know :smile_cat:

giocalitri:
:+1: after squash

PR #581, opened 2016-02-10
ab: at least one strain

### LINKED ISSUE #237: Strains: don't let the user proceed unless there is at least one strain
Currently, if the user doesn't define a strain, the user ends up with no rows in the protocols form.

annagav:
Fix #237


giocalitri:
Code looks good: will test functionality

giocalitri:
good to merge :+1:

PR #587, opened 2016-02-16
Added installation instructions.

justinabrahms:


pdpinch:
This is the same as what's in https://github.com/starteam/starcellbio_html/blob/develop/docs/DeveloperGuide.rst yes?

(with the added note that the debian install takes a long time) 

In any case, it's good to have them here, for consistency. Maybe we should add a link to the full docs.

justinabrahms:
Yes. Updated.

pdpinch:
:+1:

PR #593, opened 2016-02-22
Added support for selenium tests

### LINKED ISSUE #34: integration testing
A glance at the tests suggests that there's cleanup needed in this area. All but one test file contains a "hello world":
- https://github.com/starteam/starcellbio_html/blob/41cc4078e328e673d2cd40093bab05ee75e21d03/backend/tests.py
- https://github.com/starteam/starcellbio_html/blob/41cc4078e328e673d2cd40093bab05ee75e21d03/auth/tests.py
- https://github.com/starteam/starcellbio_html/blob/41cc4078e328e673d2cd40093bab05ee75e21d03/instructor/tests.py

And the file that contains tests is in a djangoapp, frontend_tests, which appears to be for integration testing with Selenium.
It contains four tests; here are some notes about them:
- The first test points the Chrome webdriver to `http://localhost:8000/static/index.html` and checks if the response has a `div` with an id `main`.
- This is a baked URL; it's not reverse routed.
- It doesn't appear to launch the server for you.
- This is how it initializes the webdriver: `self.driver = webdriver.Chrome(settings.rel('../../mac/chromedriver'))`
- The second test does the same thing and then checks if some element has a class indicating you're on the homepage.
- The third test does everything else: https://github.com/starteam/starcellbio_html/blob/41cc4078e328e673d2cd40093bab05ee75e21d03/frontend_tests/tests.py#L52-L252
- The last test has a bunch of commented code and executes some JavaScript. It calls the simulateDragSortable plugin on what appears to be some "western blot samples", but doesn't actually assert anything:

``` javascript
script = "$($('.scb_s_western_blot_choose_samples_list_item')[0]).simulateDragSortable({ move: 1 });"
success = self.driver.execute_script(script)
```

There are no unit tests.

justinabrahms:

Fixes #34 
#### What's this PR do?

Adds selenium testing to the build. These check all 3 experiment types: flow cytometry, microscopy, & western blotting.

tests.py  (probably easier to ignore the deletions and look only at the whole file)

See updates to DeveloperGuide.rst

Changes to build.py were made so we didn't have to do a bunch of cd-ing around in the build scripts.
#### Screenshots (if appropriate)

N/A

### DIFF
--- a/StarCellBio/settings.py
+++ b/StarCellBio/settings.py
@@ -40,7 +40,7 @@ DB_PORT = ''
 
 # Default logging configuration
 LOG_LEVEL = 'DEBUG'
-DJANGO_LOG_LEVEL = 'DEBUG'
+DJANGO_LOG_LEVEL = 'WARNING'
 # For logging to a syslog host
 LOG_HOST = 'localhost'
 LOG_HOST_PORT = 514
@@ -260,6 +260,10 @@ LOGGING = {
     },
     'loggers': {
         'root': {
+            'handlers': ['console', 'syslog'],
+            'level': LOG_LEVEL,
+        },
+        'StarCellBio': {
             'handlers': ['console', 'syslog'],
             'level': LOG_LEVEL,
         },
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,32 @@
 language: python
 python:
   - "2.7"
+cache: pip
 services:
   - mysql
-before_install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -qq libmysqlclient-dev mysql-client build-essential libncurses5-dev openjdk-6-jre npm nodejs git postfix phantomjs
+env:
+  global:
+  - ARTIFACTS_BUCKET=mitodl-travis-ci-screenshots
+  - SELENIUM_URL="http://localhost:4444/wd/hub"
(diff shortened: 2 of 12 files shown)


StarCellBio/settings.py
```diff
@@ -40,7 +40,7 @@
 
 # Default logging configuration
 LOG_LEVEL = 'DEBUG'
-DJANGO_LOG_LEVEL = 'DEBUG'
+DJANGO_LOG_LEVEL = 'WARNING'
```
blarghmatey, line 43: This is safe to leave at debug because it gets overridden during deployment by Ansible.

justinabrahms, line 43: It makes development a giant pain, because it prints the sql for every
single django query on page load.

requirements.txt
```diff
@@ -22,3 +22,4 @@ docopt>=0.6.2
 ordereddict>=1.1
 sh>=1.11
 sqlparse
+bok-choy==0.4.10
```
blarghmatey: This should probably go in `requirements.dev.txt`

StarCellBio/settings.py
blarghmatey, line 43: Fair enough, just making sure it wasn't for deployment reasons.

blarghmatey:
:+1:

requirements.txt
justinabrahms: Done.

PR #600, opened 2016-03-01
ab: enable techniques independently

### LINKED ISSUE #580: AB side menu: enable techniques independently of each other
Should be able to define flowcytometry technique before western blot.

### LINKED ISSUE #585: AB side menu: page numbering for enabling/disabling pages
Enable techniques (page links on the side menu and corresponding 'continue' buttons) independently of each other. If western blot is not selected should be able to proceed to other techniques.

annagav:

Fix #585 
Fix #580 
#### What's this PR do?

Enables techniques independently of each other.

Create a new assignment, go through all steps of defining the experiment setup. On the 'select technique' page play with various selections, try to go to facs 'Sample Prep'  without defining western blot.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -20,8 +20,9 @@ from functools import wraps
 page_order = [
     'assignment', 'course', 'strains', 'variables', 'treatments', 'protocols',
     'techniques', 'wb_lysate_type', 'wb_antibody', 'wb_band_size',
-    'wb_band_intensity', 'facs_sample_prep', 'facs_analyze'
+    'wb_band_intensity'
 ]
+facs_pages = ['facs_sample_prep', 'facs_analyze']
 
 
 @login_required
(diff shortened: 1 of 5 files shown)


instructor/common.py
```diff
@@ -85,6 +86,28 @@ def create_new_assignment(request):
...
 
 
+def get_pages(assignment):
+    """
+    Return a dictionary of enabled/disabled pages
+    """
+    pages = {}
+    enabled = True
+    for page in page_order:
+        pages[page] = enabled
```
justinabrahms, line 96: `pages[page] = page != assignment.last_page_name`

Here and below

annagav, line 96: After the case `page == assignment.last_page_name` I want `enabled` to stay False.

justinabrahms, line 96: Ahh, I see what's going on now. Thanks.

justinabrahms:
Can you write some unit tests for `get_pages`? Should just use the normal django test framework, not selenium.

The code here could use some refactoring soon-ish. You shouldn't have to have to edit so many functions with a change like this. Providing a side bar might be better as something like a django template tag, which will allow it to query for the things it needs (like page order, etc).

Not necessary for this PR, but we can talk about it tomorrow, if you want.

annagav:
Added tests for `get_pages`.

justinabrahms:
:+1:

PR #610, opened 2016-03-15
Selenium-IDE script for validating user signup

### LINKED ISSUE #606: Integration tests for creating student accounts
As a devops, I'd like a test that I can run to confirm that student accounts can be created successfully on a deployment.

For some time in production, account creation was failing with a 500 error. We're still not entirely sure what the cause was, but we'd like to have better tooling to detect it in the future.

To create a student account, we will need a course code.

justinabrahms:

Fixes #606 
#### What's this PR do?

Adds a selenium-ide script so we can test user creation on production

post-deploy-checklist, I guess, but just do the manual step below instead.

Download selenium ide for firefox. From there, you should be able to load up the file I made below, and run it against SCB.

We didn't do it as a selenium console script because iframes are terrible.


annagav:
Looks good :+1:

PR #620, opened 2016-03-25
AB facs: x axis labels

### LINKED ISSUE #591: AB: FACS allow user to change axes labels
As an instructor I would like to set the labels and ranges for the x and y axes in my histogram. 

Since the labels and ranges need to be the same for each analysis and condition, the UI should not be on the drawing interface -- which would suggest that they can be set per graph.

This should be a step after 'Sample Prep' and before adding histograms (currently 'Analyze'). The user should be able to set the minimum, maximum and any intermediate values for the x and y axes, and the text labels for each. There should also be a preview function so the instructor can see what graph labels will look like before continuing. 

The instructor will also have to choose whether she wants a linear or logarithmic axis. This will affect which pre-defined histograms she will see in the select histogram interface #574)

annagav:

Fix #591 
#### What's this PR do?

Take range and tick values for x and display it in the sketching tool.

Play with the upper bound on x and the tick values, and see how that is reflected in the sketching canvas.

The scale is only linear for now.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -1322,9 +1322,7 @@ def facs_setup(request):
     pk = request.session['assignment_id']
     assignment = models.Assignment.objects.get(id=pk)
 
-    facs, created = models.FlowCytometry.objects.get_or_create(
-        assignment=assignment
-    )
+    facs, _ = models.FlowCytometry.objects.get_or_create(assignment=assignment)
     FlowCytometryForm = modelform_factory(
         models.FlowCytometry,
         exclude=['assignment']
@@ -1471,6 +1469,7 @@ def facs_analyze(request):
     """
     pk = request.session['assignment_id']
     assignment = models.Assignment.objects.get(id=pk)
+    facs, _ = models.FlowCytometry.objects.get_or_create(assignment=assignment)
 
     instances = models.FlowCytometryHistogram.objects.filter(
         sample_prep__assignment=assignment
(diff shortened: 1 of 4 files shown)


annagav:
@giocalitri, this is rebased and ready for review.

instructor/common.py
```diff
@@ -1471,6 +1471,9 @@ def facs_analyze(request):
     """
     pk = request.session['assignment_id']
     assignment = models.Assignment.objects.get(id=pk)
+    facs, created = models.FlowCytometry.objects.get_or_create(
```
giocalitri: are you using the variable `created`?

annagav: not really

instructor/ui/sketch_tool.js
```diff
@@ -38,6 +43,21 @@ function load_sketch_tool() {
         visible: false
     });
 
+    /* Calculating axis labels */
+    tick_values=tick_values.split(",").map(function(x){
+        return parseInt(x)
+    });
+    x_upper_bound = parseInt(x_upper_bound);
+    _.each(tick_values, function(value, index){
+        var x_coor = X_AXIS_LENGTH_PX*value/x_upper_bound;
```
giocalitri: nit: please put some spaces around the operators

instructor/common.py
giocalitri: well, you should write something like this then:

```
facs, _ = models.FlowCytometry.objects.get_or_create
```

unless you use `_` for something else

annagav:
@giocalitri, comments addressed.

giocalitri:
:+1:  after squash

PR #630, opened 2016-04-08
ab facs: data points conversion

annagav:
#### What's this PR do?

When saving the drawn function, convert data points to regular coordinate system with origin at bottom left. When retrieving data convert it back for Canvas. 
Fixed the alignment of the pencil icon.

Notice how old previously drawn functions are inverted.

### DIFF
--- a/instructor/ui/sketch_tool.js
+++ b/instructor/ui/sketch_tool.js
@@ -156,3 +156,20 @@ function printText(x, y, value){
     text_obj.position.x -= text_obj.bounds.width / 2;
     return text_obj;
 }
+
+/* Translate data points to normal coordinate system */
+function getDataPoints(){
+    var correct_points = [];
+    _.each(data_points, function(point) {
+        correct_points.push([point[0] - X_ORIGIN , Y_ORIGIN - point[1]]);
+    });
+    return correct_points;
+}
+/*Converts normal points to Canvas's point of reference
+* where origin is at top-left */
+function convertToCanvas(data){
+    _.each(data, function(point){
+      point[0] += X_ORIGIN;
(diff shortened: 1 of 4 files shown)


pdpinch:
@annagav I don't think I understand how to test this. How/when is the data retrieved for rendering on the canvas? I'm testing with the develop branch now and I can't reproduce the issue of "old previously drawn functions are inverted"

annagav:
You would not be able to see any inverted functions on develop. When you are on this branch any functions that were created in the past ( on a different branch) will now look inverted.

pdpinch:
How do you think I should test this?

annagav:
Sketch graphs using the sketching tool and make sure they look normal. Only the previously created graphs on a different branch will be inverted (if you have any).

annagav:
@noisecapella, Could you please review this PR? It is blocking my work....

instructor/ui/sketch_tool.js
```diff
@@ -156,3 +156,20 @@ function printText(x, y, value){
...
+function getDataPoints(){
+    var correct_points = [];
+    _.each(data_points, function(point) {
+        correct_points.push([point[0] - X_ORIGIN , Y_ORIGIN - point[1]]);
+    });
+    return correct_points;
+}
+/*Converts normal points to Canvas's point of reference
+* where origin is at top-left */
+function convertToCanvas(data){
```
noisecapella: It's generally good practice to return a copy rather than mutate the parameter argument. This makes the function and its arguments easier to reason about. Can you change this to return a copy of `data`, maybe using `_.map` below?

annagav: Fixed!

noisecapella:
:+1:

PR #635, opened 2016-05-20
ab: change page name histogram

### LINKED ISSUE #623: FACS: change "canvas" step to "Histogram"
This is the form where users configure their axes, i.e. whether they are linear or logarithmic, the maximum value and the tick values. 

Also, the blue text at the top of the form should be updated to use the same name.

screen shot 2016-03-31 at 12 12 00 pm

annagav:

Fix #623


pdpinch:
LGTM!

👍

PR #636, opened 2016-05-23
ab facs: default condition names

### LINKED ISSUE #602: FACS: add default value for condition name
As an instructor, I may not know what "condition name" means. A default value would help me understand the meaning of this field. Ideally, the default value would change when I change the analysis to suggest a potentially relevant name. 

The term "Condition" may not be intuitive to instructors.

annagav:

Fix #602
#### What's this PR do?

Display default condition name when analysis type was selected.

FACS Sample prep page in the assignment builder.


noisecapella:
:+1:

PR #642, opened 2016-06-22
Added noise preview for newly sketched graph

### LINKED ISSUE #605: AB FACS: preview option for graphs
As an instructor, I want to be able to see what my graphs will look like for students, and I want them to look realistic. 

We will need a function to add noise to the hand-drawn graphs. 

Ideally, we would apply the function as soon as the drawing is finished. It's okay if the noise is random and slightly different every time. 
- [ ] We need to add a note to the instructor that (sample) noise is added to the graph, to make it look more realistic.

annagav:

Fix #605 
#### What's this PR do?

Add a preview with noise in the histogram tools window.

### DIFF
--- a/instructor/ui/constants.js
+++ b/instructor/ui/constants.js
@@ -13,3 +13,18 @@ Y_AXIS_LENGTH_PX = Y_ORIGIN - Y_AXIS_LAST_VALUE;
 
 Y_AXIS_TICKS = [20, 40, 60, 80];
 Y_AXIS_LENGTH_VALUE = 100;
+
+CANVAS_NOISE_WIDTH = 450;
+CANVAS_NOISE_HEIGHT = 250;
+
+CANVAS_SKETCH_WIDTH = 600;
+CANVAS_SKETCH_HEIGHT = 300;
+
+function addNoise(points, factor) {
+  var factor = factor || 0.05;
+  var noisy_points = [];
+  _.each(points, function (point) {
+    noisy_points.push([point[0], point[1] * (1 - factor + 2 * factor * Math.random())]);
+  });
+  return noisy_points;
+}
--- a/instructor/ui/instructor.js
+++ b/instructor/ui/instructor.js
@@ -382,34 +382,8 @@ $(function() {
     });
   });
 
-  $(".add_histogram_btn").click(function () {
-    /* this btn has the id of the corresponding row */
-    var row_id = $(this).data('row_id');
-    /* Get name of the sample from the row itself */
-    var sample_name = $("#" + row_id).text().replace(/(\n *)+/g, "");
-    /* cell_treatment, analysis, condition */
-    var sample_treatment  = $(this).data('sample_treatment');
-    var sample_treatment_array = sample_treatment.split(',');
-    var instance_pk = row_id.match(/(\d+)$/)[0];
-    $('.scb_ab_f_sample_name').text(sample_name);
-    $('.scb_ab_f_treatment_text').text(sample_treatment);
-    $('.scb_ab_s_histogram_dialog')
-      .data({'cell_treatment': sample_treatment_array[0], 'pk': instance_pk})
-      .css('visibility', 'visible');
-    /* Label x axis with condition name */
-    nameXAxis(sample_treatment_array[2]);
 
 
-  });
-
-  /* Close draw histogram dialog */
-  $('.scb_ab_f_close_dialog').click(function () {
-    $('.scb_ab_s_histogram_dialog').css('visibility', 'hidden');
-    if (undostack.length > 0) {
-      undostack.pop().remove();
-      data_points = [];
-    }
-  });
   $(".scb_ab_s_histogram_tab_not_selected").click(function () {
     $(".scb_ab_s_draw_histogram_view").toggle();
     $(".scb_ab_s_select_histogram_view").toggle();
@@ -420,6 +394,7 @@ $(function() {
 
   $('.scb_ab_s_histogram_dialog').draggable({handle: '.scb_ab_s_dialog_title'});
 
+  /* Mark this histogram for selection from the library */
   $('.scb_ab_f_select_histogram').click(function(){
     var canvas_library = $("[id|='library']");
     _.each(canvas_library, function (canvas) {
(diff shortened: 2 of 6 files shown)


instructor/ui/sketch_tool.js
```diff
@@ -129,7 +131,8 @@ function draw_line(x1 ,y1 ,x2 ,y2, color){
 
 }
 
-function nameXAxis(condition){
+function nameXAxis(condition, this_paper){
+    paper = this_paper;
```
noisecapella: Is this meant to be a local variable? If so we should make this `var paper = this_paper;`. Otherwise this function redefines the global `paper` which is usually not a good idea

annagav: `paper` is a global variable that keeps reference to the currently active PaperScope. So here I am trying to set the PaperScope I want to work with. I found a better way to do this: `this_paper.activate()`.

noisecapella: Ah, I see. Can you move that step and `paper.view.update()` into the calling function? `nameYAxis` doesn't do this setup so it would be good to be consistent

instructor/ui/sketch_tool.js
```diff
@@ -60,7 +61,7 @@ function load_sketch_tool(x_upper_bound, tick_values) {
 
 
     function mouseDownEvent (event) {
-        path = new Path();
+        path = new paper.Path();
```
noisecapella: What's the significance of this change? Is the new `Path` tied to that paper instance?

noisecapella:
Functionality works great

instructor/ui/sketch_tool.js
annagav: No significance, forgot to remove it.

instructor/ui/instructor.js
```diff
@@ -489,10 +474,80 @@ $(function() {
...
+          );
+          var path = new Path();
+          path.strokeColor = 'black';
+          path.strokeWidth = '1';
+          var factor = 0.05;
+          var regular_points = getDataPoints();
+          _.each(data_points, function (point, i) {
+            path.add(new Point(
+              point[0] * fraction,
+              point[1] * y_fraction + regular_points[i][1]*(factor + 2 * factor * Math.random())))
```
noisecapella: Can you refactor this noise generation into its own function? Something like this would work: `var dataWithNoise = addNoise(getDataPoints())`. It would also be nice to reuse that function in the noise generation in `FACSModelFactory.js`, not sure how difficult that is though

instructor/ui/instructor.js
```diff
@@ -489,10 +474,80 @@ $(function() {
...
+      /* Can preview noise, when graph has been plotted */
+      if (typeof(data_points) != 'undefined' && data_points.length > 0) {
+        if ($(".scb_ab_s_preview_canvas_div").css('display') == 'none') {
+          $(".scb_ab_s_preview_canvas_div").show();
+          $(".scb_ab_f_add_noise").text('CLOSE');
+          mypapers[3].activate();
+          paper.project.clear();
+          paper.project.view.viewSize.width = 450;
+          paper.project.view.viewSize.height = 250;
+          var fraction = 450.0 / 600;
```
noisecapella: Can you rename to `x_fraction`?

instructor/ui/instructor.js
```diff
@@ -489,10 +474,80 @@ $(function() {
...
+          var y_fraction = 250.0 / 300;
```
noisecapella: What do `300` and `600` mean here and in the previous line?

instructor/ui/instructor.js
```diff
@@ -489,10 +474,80 @@ $(function() {
...
+     */
+    $('.scb_ab_f_add_noise').click(function () {
+      /* Can preview noise, when graph has been plotted */
+      if (typeof(data_points) != 'undefined' && data_points.length > 0) {
+        if ($(".scb_ab_s_preview_canvas_div").css('display') == 'none') {
+          $(".scb_ab_s_preview_canvas_div").show();
+          $(".scb_ab_f_add_noise").text('CLOSE');
+          mypapers[3].activate();
+          paper.project.clear();
+          paper.project.view.viewSize.width = 450;
```
noisecapella: Can you create variables for `450` and `250` and reuse them below when defining `fraction` and `y_fraction`?

instructor/ui/sketch_tool.js
annagav: sure

instructor/ui/constants.js
```diff
@@ -12,4 +12,19 @@ X_AXIS_LENGTH_PX = X_AXIS_LAST_VALUE - X_ORIGIN;
 Y_AXIS_LENGTH_PX = Y_ORIGIN - Y_AXIS_LAST_VALUE;
 
 Y_AXIS_TICKS = [20, 40, 60, 80];
-Y_AXIS_LENGTH_VALUE = 100;
\ No newline at end of file
+Y_AXIS_LENGTH_VALUE = 100;
+
+CANVAS_NOISE_WIDTH = 450;
+CANVAS_NOISE_HEIGHT = 250;
```
pdpinch, line 18: Are these the dimensions of the canvas for noise preview?

instructor/ui/instructor.js
```diff
@@ -444,6 +419,12 @@ $(function() {
 
     var instance_id, path, data, row_id, canvas_data;
 
+    mypapers = [];
+    mypapers[0] = new paper.PaperScope();
+    mypapers[1] = new paper.PaperScope();
+    mypapers[2] = new paper.PaperScope();
+    mypapers[3] = new paper.PaperScope();
```
pdpinch, line 426: Why four of these?

instructor/ui/constants.js
annagav, line 18: Yes

instructor/ui/instructor.js
annagav, line 426: We have a total of four paper views, including the small thumbnail graphs and the library of all graphs.

instructor/ui/instructor.js
```diff
@@ -498,9 +483,78 @@ $(function() {
...
     load_sketch_tool(x_upper_bound, tick_values);
+
+
+    mypapers[3].setup("noise-canvas");
+
+    /*
+      Preview noise for sketched graph
+     */
+    $('.scb_ab_f_add_noise').click(function () {
+      console.log(mypapers[2]);
```
noisecapella: Can you remove this line?

noisecapella:
:+1:

PR #648, opened 2016-06-28
Remove Live from exercise 2

annagav:

None
#### What's this PR do?

Remove the Live treatment in `facs_kinds`, which was added to test the functionality. Now can use this functionality through defining Live and Fixed treatments in FACS in the Assignment Builder. 

To test this: perform the experiment (NSF Exercise 2), notice only `fixed` is available for cell treatment. In analyze step each sample should produce a graph.


noisecapella:
:+1:

Issue #619, opened 2016-03-24
Microscopy: discovery on file upload


pdpinch:
Research django best practices for image upload and forms

pdpinch:
@annagav it sounds like this is done?

annagav:
It is done on the development side. Opened related issue #658.

Issue #202, opened 2015-05-20
As instructor, I can upload microscopy images corresponding to specific analysis/conditions


pdpinch:
> In this page the instructor will upload the image(s) associated with each sample and its corresponding microscopy analysis/conditions. Once the instructor clicks “Add Image(s)” a pop up window will appear where the instructor can select an image from his/her computer. (see the StarCellBio AssignmentBuilder – Microscopy Page 4). Once an image is selected/uploaded the instructor can apply the image(s) to other samples. For any given image the instructor has the option of specifying the microscope objective used to obtain the image.
> 
> starcellbio_html repository docs/problemsetbuilder/User Interface Design Documents - Jamie Waters/SCB_ProblemSetBuilder_09_Microscopy.pdf  the question text on page 4 in 'SCB_ProblemSetBuilder_09_Microscopy.pdf has been updated. For the updated question wording, please see the design sketch document, 'Instructor Account UI - Microscopy Ver5.pdf'
> 
> starcellbio_html repository docs/problemsetbuilder/User Interface Sketches/Microscopy/Instructor Account UI - Microscopy Ver5.pdf  notes on the functionality required in each step to set up the microscopy experimental technique. Notes on each step are in blue text. However, design elements and layout have been modified by the graphic/user interface designer. In these cases, defer to the graphic designer document, listed above.
> 
> The graphic designer did not detail each step of setting up a microscopy experimental technique. In total there are 3 questions/steps to set up a microscopy experimental technique, and the designer sketched out steps 2 and 3. For step 1 (question wording and answer choices), please see the design sketch document, 'Instructor Account UI - Microscopy Ver5.pdf'.

pdpinch:
@annagav after #652 and #657 merged, what more needs to be done here? (It seems like this story is maybe too big)

annagav:
Copy images to other samples would be another functionality. But could be in different issue.

pdpinch:
Closing this. Remaining work is now in #703

PR #657, opened 2016-07-15
ab: microscopy image dialog

annagav:

Related to #202, #619
#### What's this PR do?

In the analyze view of microscopy setup in the assignment builder, added functionality to 'ADD IMAGE' button. The button open a dialog in which the user can upload images, then select them and save the for that particular sample.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -1344,10 +1344,10 @@ def microscopy_analyze(request):
     if request.method == "POST" and 'upload' in request.POST:
         if len(request.FILES) > 0:
             uploaded_file = request.FILES['file']
-
-            new_image = models.MicroscopyImage(file=uploaded_file)
-            new_image.assignment = assignment
-            new_image.save()
+            new_image, _ = models.MicroscopyImage.objects.get_or_create(
+                file=uploaded_file,
+                assignment=assignment
+            )
         # need few variables to keep the dialog open
         dialog_open = True
         if 'mapping_pk' in request.POST:
@@ -1386,7 +1386,7 @@ def microscopy_analyze(request):
             )
         )
 
-    all_images = models.MicroscopyImage.objects.filter()
+    all_images = models.MicroscopyImage.objects.filter(assignment=assignment)
     ImageForm = modelform_factory(models.MicroscopyImage, fields=['file'])
     image_form = ImageForm()
     variables = {
@@ -1698,6 +1698,8 @@ def facs_histograms_edit(request, assignment, sample_prep, sp):
     )
 
 
+@assignment_selected
+@check_assignment_owner
 @login_required
 def select_images(request):
     mapping_id = request.POST.get('mapping_pk')
@@ -1712,6 +1714,8 @@ def select_images(request):
     return HttpResponse('complete')
 
 
+@assignment_selected
+@check_assignment_owner
 @login_required
 def submit_histogram(request):
     """
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -17,6 +17,7 @@ from instructor.compiler import get_protocol_headers
 from django.http import HttpResponse, HttpResponseBadRequest
 from functools import wraps
 
+from django.conf import settings
 page_order = (
     'assignment', 'course', 'strains', 'variables', 'treatments', 'protocols',
     'techniques', 'wb_lysate_type', 'wb_antibody', 'wb_band_size',
@@ -1334,6 +1335,25 @@ def create_image_mappings(assignment, sample):
 def microscopy_analyze(request):
     pk = request.session['assignment_id']
     assignment = models.Assignment.objects.get(id=pk)
+    chosen_sampleprep = ""
+    chosen_protocol = ""
+    mapping_pk = ""
+    dialog_open = False
+    if request.method == "POST" and 'continue' in request.POST:
+        return redirect('common_assignments')
+    if request.method == "POST" and 'upload' in request.POST:
+        if len(request.FILES) > 0:
+            uploaded_file = request.FILES['file']
+
+            new_image = models.MicroscopyImage(file=uploaded_file)
+            new_image.assignment = assignment
+            new_image.save()
+        # need few variables to keep the dialog open
+        dialog_open = True
+        if 'mapping_pk' in request.POST:
+            chosen_protocol = request.POST.get('protocol')
+            chosen_sampleprep = request.POST.get('sample_prep')
+            mapping_pk = request.POST.get('mapping_pk')
 
     image_mappings = models.MicroscopyImageMapping.objects.filter(
         sample_prep__assignment=assignment
@@ -1345,9 +1365,6 @@ def microscopy_analyze(request):
         'strain_protocol__treatment__temperature__degrees',
         'strain_protocol__treatment__collection_time__time'
     )
-
-    if request.method == "POST" and 'continue' in request.POST:
-        return redirect('common_assignments')
 
     # Grouping ImageMapping objects by analysis and condition
     grouped_images = []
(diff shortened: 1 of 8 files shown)


noisecapella:
@annagav Can you create an issue to cover uploading images to S3 in production? cc @bdero @blarghmatey

annagav:
Opened #658

noisecapella:
When I click 'Save and Continue' on Microscopy -> Sample Prep to go to Microscopy -> Analyze, I get this error:
screenshot from 2016-07-15 11-21-19

instructor/common.py
```diff
@@ -1334,6 +1335,25 @@ def create_image_mappings(assignment, sample):
...
+    chosen_protocol = ""
+    mapping_pk = ""
+    dialog_open = False
+    if request.method == "POST" and 'continue' in request.POST:
+        return redirect('common_assignments')
+    if request.method == "POST" and 'upload' in request.POST:
+        if len(request.FILES) > 0:
+            uploaded_file = request.FILES['file']
+
+            new_image = models.MicroscopyImage(file=uploaded_file)
```
noisecapella: Any reason not to do `MicroscopyImage.objects.create(file=uploaded_file, assignment=assignment)` instead?

instructor/common.py
```diff
@@ -1369,6 +1386,9 @@ def microscopy_analyze(request):
             )
         )
 
+    all_images = models.MicroscopyImage.objects.filter()
```
noisecapella: Can you use `all()` instead to be explicit?

instructor/common.py
```diff
@@ -1671,6 +1698,20 @@ def facs_histograms_edit(request, assignment, sample_prep, sp):
...
+        selected_image = get_object_or_404(models.MicroscopyImage, pk=image_pk)
+
+        instance.images.add(selected_image)
+    instance.save()
+
+    return HttpResponse('complete')
+
+
+@login_required
 def submit_histogram(request):
```
noisecapella, line 1727: Do you need the `@assignment_selected` and `@check_assignment_owner` decorators here and on `select_images`?

instructor/common.py
```diff
@@ -1671,6 +1698,20 @@ def facs_histograms_edit(request, assignment, sample_prep, sp):
 
 
 @login_required
+def select_images(request):
+    mapping_id = request.POST.get('mapping_pk')
+    image_pk_list = request.POST.getlist('image_pk_list[]')
```
noisecapella, line 1705: I don't think you need the `[]` here

instructor/ui/instructor.js
```diff
@@ -378,7 +378,22 @@ $(function() {
...
-
+  $('.scb_ab_f_save_image').click(function(){
+    var data={};
+    data['mapping_pk'] = $(this).data('pk');
+    data['image_pk_list'] = _.map($('.scb_ab_s_image_selected'), function(element){
+      return $(element).attr('id').match(/(\d+)$/)[0];
+    });
+    if(data['image_pk_list'].length>0){
+      $.ajax({
+      url: '/ab/assignments/select_images/',
```
noisecapella: Can you indent the other lines in the `if` block here?

instructor/ui/instructor.js
```diff
@@ -542,14 +566,53 @@ $(function() {
...
     nameXAxis(sample_treatment_array[2]);
   });
 
+  $(".open_upload_window_btn").click(function(){
+ /* this btn has the id of the corresponding row */
+    var row_id = $(this).data('row_id');
+    /* Get name of the sample from the row itself */
+    var sample_name = $("#" + row_id).text().replace(/(\n *)+/g, "");
+    $('.scb_ab_f_sample_name').text(sample_name);
+    var instance_pk = row_id.match(/(\d+)$/)[0];
```
noisecapella, line 590: Is the `match` necessary here? Is `row_id` ever not a number?

instructor/common.py
annagav: oops, need to filter by assignment.

instructor/common.py
annagav: no reason, I'll change to `get_or_create` to avoid upload of the same image more than once.

instructor/common.py
annagav, line 1727: Hmmm, yes I think so....

instructor/common.py
annagav, line 1705: This is how the attribute is passed in request.POST.
`request.POST.getlist('image_pk_list')` gives empty list.

instructor/ui/instructor.js
annagav, line 590: It is a string ending with a number.

instructor/ui/instructor.js
annagav: sure

instructor/common.py
noisecapella: I'm not sure `get` works with `FileField` but I don't have time to research now. I think this is fine though it won't prevent duplicate images

noisecapella:
After I upload an image the word `true` is added to the title:

screenshot from 2016-08-01 11-41-58

instructor/common.py
```diff
@@ -1380,6 +1400,13 @@ def microscopy_analyze(request):
         'instructor/micro_analyze.html',
         {
             'access': json.dumps(assignment.access),
+            'all_images': all_images,
+            'MEDIA_ROOT': settings.MEDIA_ROOT,
```
noisecapella: Can you remove this line, if we're not using it anywhere?

instructor/common.py
```diff
@@ -1670,6 +1697,24 @@ def facs_histograms_edit(request, assignment, sample_prep, sp):
...
 
+@assignment_selected
+@check_assignment_owner
+@login_required
+def select_images(request):
+    mapping_id = request.POST.get('mapping_pk')
+    image_pk_list = request.POST.getlist('image_pk_list[]')
+    instance = get_object_or_404(models.MicroscopyImageMapping, pk=mapping_id)
+    for image_pk in image_pk_list:
+        selected_image = get_object_or_404(models.MicroscopyImage, pk=image_pk)
```
noisecapella: Do we need to also filter by assignment id here? Otherwise the user can pass in arbitrary id numbers for assignments which aren't theirs

instructor/common.py
```diff
@@ -1670,6 +1697,24 @@ def facs_histograms_edit(request, assignment, sample_prep, sp):
...
+
+        instance.images.add(selected_image)
+    instance.save()
+
+    return HttpResponse('complete')
```
noisecapella: We should not return anything for content here unless it has meaning to a client. Here it seems like the default `200` status code is enough so we should return `HttpResponse()` instead

instructor/common.py
```diff
@@ -1670,6 +1697,24 @@ def facs_histograms_edit(request, assignment, sample_prep, sp):
...
+
+
+@assignment_selected
+@check_assignment_owner
 @login_required
 def submit_histogram(request):
     """
```
noisecapella, line 1728: In the lines below I think we need to filter `mapping_id` and `histogram_pk` to also filter by `assignment_id` so the user can't provide id numbers which don't match their assignment id

instructor/common.py
```diff
@@ -1670,6 +1697,24 @@ def facs_histograms_edit(request, assignment, sample_prep, sp):
     )
 
 
+@assignment_selected
+@check_assignment_owner
+@login_required
+def select_images(request):
+    mapping_id = request.POST.get('mapping_pk')
+    image_pk_list = request.POST.getlist('image_pk_list[]')
+    instance = get_object_or_404(models.MicroscopyImageMapping, pk=mapping_id)
```
noisecapella: We should filter by assignment id here too

instructor/common.py
annagav: done :)

instructor/common.py
annagav: adding it

instructor/common.py
annagav: sounds good

instructor/common.py
annagav, line 1728: fixed

annagav:
@noisecapella, I addressed all comments.

noisecapella:
Looks great :+1:

PR #662, opened 2016-07-20
ab facs: copy histogram functionality

### LINKED ISSUE #655: FACS: instructor can copy histogram from one condition to another
As an instructor, I may want to use the same histogram for multiple combinations of samples and conditions. I'd like to have an easy way to copy the same histogram from other conditions. 

There's a design for this, but it uses a checkbox to initiate the action of choosing conditions to copy. This should be a button, labelled copy instead of a checkbox.

### LINKED ISSUE #667: AB FACS: disable edit and copy buttons for samples without histograms
As an instructor, I don't expect to be able to edit or copy empty histograms.

annagav:

Fix #655 
Partially fixes #667 
#### What's this PR do?

Add functionality to copy a histogram from my sample to others.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -1628,7 +1628,7 @@ def facs_analyze(request):
     all_histograms = models.FlowCytometryHistogram.objects.exclude(
         data__isnull=True
     )
-    all_histogram_mapping = {
+    all_histogram_data = {
         int(instance.id): instance.data
         for instance in all_histograms
     }
@@ -1644,7 +1644,7 @@ def facs_analyze(request):
         'instructor/facs_analyze.html',
         {
             'histograms': json.dumps(selected_histograms_mapping),
-            'all_histograms_mapping': json.dumps(all_histogram_mapping),
+            'all_histograms_data': json.dumps(all_histogram_data),
             'all_histograms': all_histograms,
             'access': json.dumps(assignment.access),
             'x_upper_bound': json.dumps(facs.xrange),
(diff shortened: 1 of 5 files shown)


instructor/common.py
```diff
@@ -1670,6 +1670,40 @@ def facs_histograms_edit(request, assignment, sample_prep, sp):
...
+        mapping = get_object_or_404(
+            models.FlowCytometryHistogramMapping,
+            id=sample['id']
+        )
+        if sample['cell_treatment'] == 'Live':
+            mapping.live_data = histogram
+        elif sample['cell_treatment'] == 'Fixed':
+            mapping.fixed_data = histogram
+        mapping.save()
+    return HttpResponse('complete')
```
noisecapella: Can you remove `'complete'` here?

instructor/common.py
```diff
@@ -1670,6 +1670,40 @@ def facs_histograms_edit(request, assignment, sample_prep, sp):
...
+@check_assignment_owner
+@login_required
+def copy_histogram(request):
+    """
+    Copy a histogram from one mapping to a given list mappings
+    """
+    mapping_id = request.POST.get('copy_from_pk')
+    cell_treatment = request.POST.get('cell_treatment')
+    copy_to_list = json.loads(request.POST.get('copy_to_list'))
+    copy_from_mapping = get_object_or_404(
```
noisecapella, line 1735: We should also filter by assignment id here to make sure the user doesn't access data outside their assignment

instructor/common.py
```diff
@@ -1670,6 +1670,40 @@ def facs_histograms_edit(request, assignment, sample_prep, sp):
...
+
+    if cell_treatment == 'Live':
+        histogram = copy_from_mapping.live_data
+    elif cell_treatment == 'Fixed':
+        histogram = copy_from_mapping.fixed_data
+
+    for sample in copy_to_list:
+        mapping = get_object_or_404(
+            models.FlowCytometryHistogramMapping,
+            id=sample['id']
```
noisecapella: We should filter by assignment id here

noisecapella:
The issue says that the checkbox should be a button saying 'Copy' instead

noisecapella:
Functionality looks good!

annagav:
Rebased and addressed comments.

instructor/common.py
annagav: done

instructor/common.py
annagav, line 1735: adding filter

noisecapella:
:+1:

Issue #674, opened 2016-07-28
AB: 'Save and Finish' button should validate assignment and show a message.


annagav:

pdpinch:
@annagav @laleman did Racoon Gang take care of this?

PR #698, opened 2016-08-12
ab: save and continue button

annagav:

Related to #674 
#### What's this PR do?

When on the last page of Western Blot, the button should be 'save and continue' if there are more techniques enabled in the 'Select Technique' page ( and should lead to next enabled technique).
When on the last page of Microscopy, the button should be 'save and continue' and lead to 'Sample Prep' page of facs if facs is enabled in this assignment.


pdpinch:
LGTM, once tests pass

pdpinch:
👍

PR #700, opened 2016-08-12
ab: disable wb page links

### LINKED ISSUE #699: AB: disable Western Blot pages when WB is not selected on 'Select Techniques' page

annagav:

Fix #699

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -98,6 +98,11 @@ def get_pages(assignment):
         if page == assignment.last_page_name:
             enabled = False
 
+    if not assignment.has_wb:
+        wb_pages = page_order[-4:]
+        for page in wb_pages:
+            pages[page] = False
+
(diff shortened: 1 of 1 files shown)


instructor/common.py
```diff
@@ -98,6 +98,11 @@ def get_pages(assignment):
         if page == assignment.last_page_name:
             enabled = False
 
+    if not assignment.has_wb:
+        wb_pages = page_order[-4:]
```
pdpinch, line 102: This works, but I'm curious why you didn't factor the wb pages our into their own list, like micro_pages and facs_pages

pdpinch:
I have a question, but otherwise this gets a :+1: after it's rebased and the tests pass.

instructor/common.py
annagav, line 102: This approach took less changes, given the original implementation.

PR #702, opened 2016-08-12
ab: strains fixes

### LINKED ISSUE #701: AB Strains: enable save and continue button on input
If there are no strains defined for experimentation the 'continue' button is disabled. Need to enable the 'save and continue' button when the user starts typing into the field.

annagav:

Fix #701 
#### What's this PR do?

Enable 'save and continue button when the user starts typing into the text field. 
Fixes the bug where an extra form appears, when user entered first strain and clicked 'Save'.


pdpinch:
Nice! 👍 after rebase and tests pass.

PR #705, opened 2016-08-17
ab micro: implemented filters

### LINKED ISSUE #697: AB Microscopy: Allow instructor to upload images for different filters, when using Fluorescent analysis
When an instructor chooses antibody-labeling for analysis and then progresses to the upload form, she needs to be able to upload an image for each sample, with each of four filters: red, blue, green and all filters. 

As elsewhere in microscopy, not all images are required.

annagav:

Partially fixes #697
Does not include publishing of the assignment with filters.
#### What's this PR do?

Add filters to Microscopy 'Sample Prep' and 'Analyze' page.
If the analysis type is not fluorescent then the filters selection is not shown.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -16,6 +16,7 @@ import re
 from instructor.compiler import get_protocol_headers
 from django.http import HttpResponse, HttpResponseBadRequest
 from functools import wraps
+from django.template.defaulttags import register
 
 page_order = (
     'assignment', 'course', 'strains', 'variables', 'treatments', 'protocols',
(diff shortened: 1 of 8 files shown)


pdpinch:
If this does not include publishing, does that also meant that preview isn't working?

pdpinch:
I'm not sure I understand the distinction between "red" "blue" and "green" and "all" -- does "all" mean that all filters have been applied? Does it make sense to have it as a checkbox, or does choosing "all" mean that the others should be deselected?

pdpinch:
Is it possible to display the filter checkboxes when the appropriate analysis is chosen, instead of on save? It's unintuitive to have to click save in order to get the complete form.

PR #707, opened 2016-08-19
Create gen directories if not already there

noisecapella:
#### What's this PR do?

Reverts changes to #706 and replaces the fix with one that creates `gen` directories. It looks like the actual cause of the problem was that the soy generation created `gen` directories automatically but the gss generation did not


noisecapella:
Can you review this too @annagav ?

annagav:
I think it creates `gen` folders in all directories, even where there is nothing to generate.

noisecapella:
I changed it to only create the `gen` directory before the `gss` files are compiled

annagav:
👍

PR #708, opened 2016-08-19
Switched tests to use latest stable chromium

noisecapella:
#### What's this PR do?

Switched from Firefox to Chromium. This was part of the work to fix the build which ended up not being necessary but it's still a good idea since this is the only browser supported

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,15 +7,20 @@ services:
 env:
   global:
   - ARTIFACTS_BUCKET=mitodl-travis-ci-screenshots
-  - SELENIUM_URL="http://localhost:4444/wd/hub"
+  - SELENIUM_URL="http://localhost:9515"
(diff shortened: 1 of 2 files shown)


annagav:
@noisecapella, Could you please take a look at the 'Developers Guide' on selenium testing to see if it needs to be updated with these new changes?

pdpinch:
I can take a look. I already noticed a few cosmetic errors in these instructions.

noisecapella:
Sure, from a quick glance it looks we need to update:
- the port changed from 4444 to 9515 (not sure what `14444` was for)
- Instead of the seleniumdriver jar it now uses the chromedriver binary.

annagav:
👍

noisecapella:
I'll merge this then create an issue to update the developer's guide

Issue #658, opened 2016-07-15
AB Microscopy: Upload images to S3 in production.


annagav:
DevOps needs to:
- [ ] create s3 bucket
- [ ] update configuration for dev server and live server (?)

pdpinch:
@starteam/devops can you guys give us some guidance on what needs to be done with this?

ald32:
To clarify, am I part of the star team?

blarghmatey:
I believe this is complete. I am happy to reopen it if that is not the case.

PR #717, opened 2016-08-30
Setup default file storage S3

annagav:

Related to #658
#### What's this PR do?

Setup django storage to use Amazon S3 to store image files.

### DIFF
--- a/StarCellBio/settings.py
+++ b/StarCellBio/settings.py
@@ -13,6 +13,7 @@ rel = lambda p: os.path.join(SITE_ROOT, p)
 
 DEBUG = True
 TEMPLATE_DEBUG = DEBUG
+S3_BACKEND_ENABLED = False
 
 import platform
 if platform.node() == 'starapp':
@@ -25,6 +26,9 @@ else:
 ADMINS = (
     # ('Your Name', 'your_email@example.com'),
 )
+
+if S3_BACKEND_ENABLED:
+    DEFAULT_FILE_STORAGE = 'libs.storages.S3Storage.S3Storage'
 
 MANAGERS = ADMINS
 
@@ -140,6 +144,7 @@ INSTALLED_APPS = (
     'django.contrib.admindocs',
     'backend',
     'instructor',
+    'storages'
 ) + auth.settings.INSTALLED_APPS
 
 # django all-auth config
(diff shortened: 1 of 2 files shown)


StarCellBio/settings.py
```diff
@@ -26,6 +26,12 @@
     # ('Your Name', 'your_email@example.com'),
 )
 
+if os.uname()[1] != 'vagrant-ubuntu-trusty-64':
+    DEFAULT_FILE_STORAGE = 'libs.storages.S3Storage.S3Storage'
+    AWS_ACCESS_KEY_ID = ''
+    AWS_SECRET_ACCESS_KEY = ''
+    AWS_STORAGE_BUCKET_NAME = ''
```
blarghmatey: Rather than making a conditional based on the uname of the host I think it would be easier to use something like `S3_BACKEND_ENABLED = False` which can be overridden at deploy time.

StarCellBio/settings.py
```diff
@@ -140,6 +146,7 @@
     'django.contrib.admindocs',
     'backend',
     'instructor',
+    'storages'
```
blarghmatey, line 147: I'm assuming this is using the `django-storages` package. Did you add that requirement to the `requirements.txt`?

annagav: yes :)

StarCellBio/settings.py
```diff
@@ -26,6 +27,12 @@
     # ('Your Name', 'your_email@example.com'),
 )
 
+if S3_BACKEND_ENABLED:
+    DEFAULT_FILE_STORAGE = 'libs.storages.S3Storage.S3Storage'
+    AWS_ACCESS_KEY_ID = ''
+    AWS_SECRET_ACCESS_KEY = ''
+    AWS_STORAGE_BUCKET_NAME = ''
```
blarghmatey: Rather than setting the `AWS` variables to empty strings it might be better to leave them out and only set the `DEFAULT_FILE_STORAGE` when the s3 variable is set. The other settings will be added in through Ansible and I'm not sure of the order of evaluation between that and the settings here. If left as is these variables may overwrite what I am setting as part of the deployment.

annagav:
@blarghmatey, Comment addressed!

blarghmatey:
👍

PR #725, opened 2016-09-02
ab micro: image objective

### LINKED ISSUE #695: AB Microscopy: allow instructor to set a value for objective
The objective is non-functional field that indicates the magnification of a particular microscropy image. The instructor can set the value, and it's simply displayed to students during the experiment. 

This is an optional field, so the column label should be "Objective (optional)." If the instructor doesn't set a value, "N/A" is displayed. The field will have sample value of "4x"

screen shot 2016-08-11 at 2 48 18 pm

screen shot 2016-08-11 at 2 44 07 pm

annagav:

Fix #695
#### What's this PR do?

Add Objective input box when uploading image for Microscopy.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -1385,24 +1385,30 @@ def microscopy_analyze(request):
     mapping_pk = ""
     filter_group_id = ""
     dialog_open = False
-    if request.method == "POST" and 'continue' in request.POST:
-        if assignment.has_fc:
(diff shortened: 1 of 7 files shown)


noisecapella:
Functionality looks good. I filed #752 but I think it's unrelated

noisecapella (approved):
Looks great! :+1:

PR #726, opened 2016-09-06
Added boto3 for S3 storage

blarghmatey:
**Added boto3 for S3 storage**
Added boto3 library to support S3 storage in django-storages
backend. Removed S3_BACKEND_ENABLED flag as it wasn't actually firing
due to how YAML overrides are loaded.


annagav:
👍

PR #753, opened 2016-12-05
NSF#2: change order of techniques

### LINKED ISSUE #152: NSF Exercise 2 - Design page
There should be a vertical dividing line between the two techniques. Remove the vertical dividing line to the right of western blotting.

annagav:
Fix #152
#### What's this PR do?

Change the order of techniques, to fix the position of the divider on the 'Design Page'.

Before testing this, don't forget to delete assignment #2 (or all assignments) from `backend.Assignment`, only then it will create new assignments with the updated template at the time when you go to NSF exercises.


noisecapella (approved):
:+1:

PR #754, opened 2016-12-05
AB course setup: update course name/code bug

### LINKED ISSUE #250: Course Setup: consider warning user when changing course code
The course code is the primary key in the course model. So if the user changes the course code, it will change the code for all assignments associated with that course. 

Do we need to warn users when this happens?

annagav:
Fix #250
#### What's this PR do?

Fixed a bug related to saving changes to existing courses.
Added instruction warning users about the changes to courses.

screen shot 2016-12-05 at 3 31 09 pm

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -387,6 +387,8 @@ def course_modify(request):
                     course.owner = user
                     course.save()
                     new_course_pk = course.id
+                else:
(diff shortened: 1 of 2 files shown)


pdpinch:
LGTM 👍

PR #755, opened 2016-12-09
Enable errors when setting up a course

annagav:
None.

To test this, create new instructors account. Create an assignment, when setting up the course try to enter a course code that you have used in other accounts. That should show an error. 
Once you created few courses, try to delete them. Course that has assignments associated with it should show error when the user tries to delete it.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -210,9 +210,6 @@ def course_setup(request):
     if course_selected:
         return redirect("common_course_modify")
 
-    formset = CourseFormSet(
-        queryset=models.Course.objects.filter(owner=request.user)
-    )
     pages = {'assignment': True, 'course': True}
     return render_to_response(
         'instructor/course_modify.html',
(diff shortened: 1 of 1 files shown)


instructor/common.py
```diff
@@ -210,9 +210,6 @@ def course_setup(request):
     if course_selected:
         return redirect("common_course_modify")
 
-    formset = CourseFormSet(
```
noisecapella, line 213: Will `formset` ever be undefined when it's referenced below?

noisecapella (changes requested):
I'm getting this error when clicking 'Save and continue' on the assignment setup:

screenshot from 2016-12-09 11-48-42

instructor/common.py
annagav: I think I fixed it.

noisecapella (approved):
Looks great :+1:

Issue #749, opened 2016-12-01
AB Assignment Setup: Cannot create an assignment from an existing one.


annagav:

jmclaus:
Dealt with in https://github.com/starteam/starcellbio_html/pull/787

PR #760, opened 2017-03-17
Create new assignment based on another

annagav:
Related to #749
#### What's this PR do?
Allows the user to create a new assignment based on an existing one. Copies only experimental setup of the original assignment.
Try creating an assignment. In 'assignment setup' choose the base assignment. Click 'next and continue' and check for experimental setup to be filled out just like in the original assignment.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -17,6 +17,7 @@ from instructor.compiler import get_protocol_headers
 from django.http import HttpResponse, HttpResponseBadRequest
 from functools import wraps
 from django.template.defaulttags import register
+from copy import deepcopy
 
 page_order = (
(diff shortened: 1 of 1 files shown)


pdpinch:
@ald32 @laleman would jmclaus or idegtiarov be able to review this for Anna?

Or should we just close it?

jmclaus:
@pdpinch @annagav @idegtiarov @laleman 
I tested the branch locally, it works like a charm and clones properly an assignment. I then merged the branch onto latest develop, again locally. It works but some fields seem to loose their value in the clone. In any case, considering that @idegtiarov is about to merge a PR and I too (2 PRs, another one is coming shortly), I'd recommend we postpone this until the above is done.

PR #767, opened 2017-08-01
Fix LTI validation process

idegtiarov:
It was found out that LTI validation process has mixes of checking and validating methods.

This PR clarify LTI validation process and remove redundant methods.

All client_key validation process is moved to `validate_client_key` method. 
Method `check_client_key` was removed as redundant one.

In `validate_timestamp_and_nonce` is added timestamp checking that cannot be less than previous request's timestamp received from the certain LTI consumer. And nonce checking that cannot be duplicated. 

Timestamp and nonce are storing in the cache with the TIMEOUT option which is by default set to 10 seconds and it is fair enough to prevent replay-attack through the LTI.

Add README file with the link to documentation.

### DIFF
--- a/StarCellBio/settings.py
+++ b/StarCellBio/settings.py
@@ -162,6 +162,9 @@ CACHES = {
         'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
         'LOCATION': '/var/tmp/django_cache',
     },
+    # NOTE(idegtiarov) LTI request validation required nonce and timestamp caching with small TIMEOUT param.
+    # To ensure all lti related data has the same lifetime and frequently deleted items are split from other cache data
+    # 'lti_cache' was added.
     'lti_cache': {
         'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
         'LOCATION': '/var/tmp/django_lti_cache',
--- a/StarCellBio/settings.py
+++ b/StarCellBio/settings.py
@@ -161,6 +161,11 @@ CACHES = {
     'default': {
         'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
         'LOCATION': '/var/tmp/django_cache',
+    },
+    'lti_cache': {
+        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
+        'LOCATION': '/var/tmp/django_lti_cache',
+        'TIMEOUT': 10,
     }
 }
 
--- a/lti_provider/validator.py
+++ b/lti_provider/validator.py
@@ -29,6 +29,12 @@ class RequestValidator(oauth1.RequestValidator):
         """
         Validate LTI request's timestamp and nonce
 
+        Timestamp is validated to be equal or greater than the timestamp used in previous requests from certain
+        LTI Consumer.
+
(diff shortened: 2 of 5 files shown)


StarCellBio/settings.py
```diff
@@ -161,6 +161,11 @@
     'default': {
         'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
         'LOCATION': '/var/tmp/django_cache',
+    },
+    'lti_cache': {
```
auraz: Why use different cache and not the same?

lti_provider/validator.py
```diff
@@ -23,46 +25,55 @@ def enforce_ssl(self):
...
+        message = "LTI request's {} is not valid."
+
         logger.debug('Timestamp validating is started.')
-        today = datetime.date.today()
-        consumer_expired_date = self.consumer.expiration_date
-        if consumer_expired_date and consumer_expired_date < today:
-            raise oauth1.OAuth1Error('Consumer Key is expired.')
+        timestamp = int(timestamp)
+        timestamp_key = '{}_timestamp'.format(client_key)
+        cache_timestamp = self.cache.get(timestamp_key, timestamp)
```
auraz, line 49: i do not think we need to cache timestamp, we can just check that current time is within 10 seconds from timestamp. Can discuss it.

auraz (changes requested):

lti_provider/validator.py
idegtiarov: Timestamp is already checked that it is within `timestamp_lifetime` from the current time in `oauthlib.oauth1.rfc5849.endpoints.base` L167. Here we validate timestamp where validation is: 
```
The timestamp value MUST be a positive integer and 
MUST be equal or greater than the timestamp used in previous requests.
```
Taken from the documentation [link](https://oauth.net/core/1.0/#nonce)

StarCellBio/settings.py
idegtiarov: We add pretty small TIMEOUT param which makes the cache to delete overdue data.

auraz: please add comment

lti_provider/validator.py
auraz: thank you

auraz (approved):

lti_provider/validator.py
```diff
@@ -23,46 +25,55 @@ def enforce_ssl(self):
...
-        try:
-            self.consumer = Consumer.objects.get(consumer_key=client_key)
-        except Consumer.DoesNotExist:
-            raise oauth1.OAuth1Error('Consumer with the key {} is not found.'.format(client_key))
-        return super(RequestValidator, self).check_client_key(client_key)
-
     def validate_timestamp_and_nonce(self, client_key, timestamp, nonce, request):
         """
-        Validate LTI Consumer has not expired key
+        Validate LTI request's timestamp and nonce
```
auraz: update docstring

auraz (approved):

StarCellBio/settings.py
idegtiarov: done

lti_provider/validator.py
idegtiarov: done

annagav:
Looks good 👍

PR #768, opened 2017-08-07
Added allowed_hosts setting

blarghmatey:
Added the `ALLOWED_HOSTS` setting so that we stop getting erroneous emails about unallowed hosts.


shaidar:
:thumbsup:

PR #773, opened 2017-08-31
Add possibility for multiple image loading in assignment creation process

idegtiarov:
Add possibility to add multiple images during adding Microscopy Experimental Technique in Assignment creating process.


annagav (approved):
👍

PR #769, opened 2017-08-17
Remove libraries that do not comply with TLO's Open Source requirements

jmclaus:
Le JS libraries `jqdialog.js` (GPL version unclear) and `soyutils.js` (Apache 2.0) cannot be distributed along with StarCellBio to meet TLO's requirement for open-sourcing. Instead, these will be fetched on installation by the user:

- For `soyutils.js` with `npm` by installing `gulp-soy`. See discussion [here](https://github.com/starteam/starcellbio_html/pull/654).

- For `jqdialog` with Ansible from the following [GitHub repo](https://raw.githubusercontent.com/knadh/jqdialog/master/jqdialog.js).

PR #774, opened 2017-09-05
Fix tests

jmclaus:
This PR fixes 404 error when fetching `'soyutils.js` during preview template load.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,11 +23,12 @@ before_script:
 - ./chromedriver > /tmp/chromedriver.log 2> /tmp/chromedriver.err &
 - pip install -r requirements.txt
 - wget https://raw.githubusercontent.com/knadh/jqdialog/master/jqdialog.js -P html_app/js/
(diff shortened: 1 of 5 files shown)


annagav (approved):
👍

jmclaus:
@annagav Installing `npm_modules` into `html_app` was a very bad idea ([see PR](https://github.com/starteam/starcellbio_html/pull/769)). None of the numerous files that ended up populating `npm_modules` were needed except `soyutils.js`. It would slow down `collectstatic` and even choke the Travis build on certain occasions thus resulting in test failures. And it would also break the production build script as `clean-css` and `uglify-js` were not found at their past location.

This PR is now totally different. `npm_modules` is left outside of `html_app` and only `soyutils.js`  is copied to `html_app/js` after `npm install` in both `ansible-dev.yml` and `.travis.yml`. And changes to the 2 templates no longer needed.

annagav:
Works fine! 👍

PR #776, opened 2017-11-16
Add assignment text and multiple files upload.

jmclaus:
This PR lets an instructor add these optional features to the opening page of an assignment:

1. Text displayed at the top, just below the title. For the time being, no formatting at all is displayed.
2. Files of any type that can be downloaded by the student. They are displayed in a list of clickable anchors.

student

On the instructor side, on initial assignment setup, a text area with a placeholder and an 'Add files' button have been appended. The latter gives the instructor the possibility to add multiple files to the assignment that will only be submitted to the server when the [Save and Continue] button is clicked.

instructor-initial-setup

Once this is done, the assignment setup can still be modified. This time, green trash buttons appear next to each file anchors, giving the possibility to delete them. Additional files can also be added. This list will also be uploaded to the server only if [Save] or [Save and Continue] are clicked.

instructor-modify-setup

In the JS file, I aligned myself to the rather unusual convention for naming variables (using underscores) instead of the more common camelcase.

### DIFF
--- a/html_app/ui/assignments.gss
+++ b/html_app/ui/assignments.gss
@@ -151,6 +151,18 @@ background-image:url('../../images/western_blot/scb_gray_right_arrow_hover.png')
 	display:block;
 }
 
+.scb_s_display_section li {
+    margin-bottom: 0.5em;
+}
+
+.scb_s_display_section a {
+    color: #316f94;
+}
+
+.scb_s_display_section a:hover {
+    color: #676767;
+}
+
 .scb_assignments_header_link_selected{
 	color:white;
 	background-repeat:no-repeat;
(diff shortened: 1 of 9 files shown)


annagav (changes requested):
I think the links to uploaded files don't look clickable. Maybe use color or some icon to make it look like a clickable link to a document. Otherwise, works fine.

jmclaus:
@annagav Indeed, you have a point, thanks. I changed the color of the links to the same blue used in the header text but not in strong. Left the light gray hover color. Don't think we should add a color to the palette just for that.

annagav (approved):

jmclaus:
@annagav @idegtiarov I rebased this branch on develop and checked that it didn't break anything with @idegtiarov recent merge.

annagav:
👍

PR #777, opened 2017-11-18
Fix Ansible script

jmclaus:
@idegtiarov @annagav @laleman @pdpinch 

I had to do a full reinstall on my machine and found out the Ansible script would stall with latest version. Reason, the command:

 `sudo: yes`

is deprecated and should be replaced by the following combination:

```
become: true
become_method: sudo
```
Please look at [this](http://docs.ansible.com/ansible/latest/become.html) Ansible doc.


annagav (approved):
👍

PR #778, opened 2017-11-20
Improve histograms

jmclaus:
- Add upper value for y-axis and implement log scale.
- Remove x-axis ticks field. Ticks for both axis will be automatically generated.
- Remove former preview canvases. Use square canvases to match student view.
- Use polyline with draggable points to set up a guide for the noisy curve whose variation is determined by a slider
- Use 'pointer' cursor for histogram tabs.
improve-histograms-01
improve-histograms-02
improve-histograms-03


instructor/migrations/0023_flowcytometry.py
```diff
@@ -26,18 +26,13 @@ class Migration(migrations.Migration):
                         max_length=50,
                         choices=[
                             (b'linear', b'Linear'), (
-                                b'log', b'Logarithmic Scale'
+                                b'log', b'Logarithmic'
                             )
```
annagav: Why this is not a new migration to modify the fields of an existing model? Are we supposed to unapply `0023_flowcytometry`?

annagav (changes requested):

instructor/migrations/0023_flowcytometry.py
jmclaus: @annagav Good catch, thank you. It is corrected, reverted the changes to the above file and instead created a new migration called `0034_improve_histograms.py`

jmclaus:
@annagav @idegtiarov I rebased the branch on develop and checked out that it didn't break anything with @idegtiarov recent merge.

annagav:
Works great! The only thing the button doesn't fit the dialog:
screen shot 2017-11-21 at 2 47 03 pm

jmclaus:
@annagav Thanks. I fixed the button positioning and also decreased a bit the size of the div containing the histogram library so that only exactly 2 rows appear. I have scrollbars on both latest Chromium and Firefox but I do not see any in your screenshot. Do you have any?

select-histogram-popup

annagav (approved):

annagav:
@jmclaus On a Mac the scroll bars are only visible when you are scrolling. I don't think this is a problem. 
Looks good! 👍

jmclaus:
@annagav OK, great, I am on Linux. Will merge this when I get final approval from @laleman.

PR #785, opened 2017-12-19
Fix dialog overlay

jmclaus:
The `Add Samples` button displays, when clicked, a dialog with a gray background overlay that only covers part of the application:

overlay-before

This PR modifies the CSS so that it covers all the application:

overlay-after

Running `build.py` is necessary to see these changes applied.


annagav (approved):
👍

PR #786, opened 2018-01-26
Student and Instructor accessibility

jmclaus:
This PR enables focusability on all GUI elements that were formerly only clickable.

To attain this goal, all occurences of divs or spans that were used to mimic a clickable element were replaced by a corresponding button or anchor. Tabbing order was corrected to ensure a flow from top to bottom and left to right. All dialog windows now open up with focus on either their 'close' button or on the first focusable element of their body. That includes all the account management windows (Sign In, Reset Password, Create Instructor or Student Account etc.). Also, in these, the superfluous '*' character that was appended to the end of the labels was removed and error reporting now behaves correctly.

A black focus indicator was chosen and efforts were made to ensure it sticks closely to the shape of the GUI element ie it will be rounded for round buttons, rectangular for links and it will even follow the outline of some tabs were a triangle is attached below them.

In the instructor application,invisible links have been added to the right of the logo: 'Skip to sidebar' and 'Skip to content'. They will only appear if tab or shift+tab is pressed on the element preceding them (logo) or following them (contact). If enter is pressed, focus will be given to the corresponding item in the sidebar or in the content. The student application only contains a 'Skip to content' invisible link.

In every file that was going to be modified, an attempt was first made to start fixing their formatting which contained a mix or tabs and spaces, unnecessary blanks lines, erroneous indentation or missing blank line at the end of the file.

Functionality was barely touched and test always passed all along the way. A good way to test the changes is to try to create an assignment by using only the keyboard. And then visualize the result, student side, using the same method.

Here are the areas left though that are not fully keyboard accessible:

**Instructor side:**
- `Drag & Drop` in `Microscopy > Images`
- `Generating histograms` in` Flow Cytometry > Analyze`

**Student side:**
- The pages beyond `Western Blot Analyses`, `Flow Cytometry Analyses`, `Microscopy Analyses`. By inheriting some of the default CSS, most GUI elements, but not all, will display the focus indicator when tabbed to. The tabbing order might be incorrect though.

There are a lot of commits but most of them are small and can be squashed if you wish once the review is done and the PR accepted.

### DIFF
--- a/auth/templates/uniform/allauth/account/login.html
+++ b/auth/templates/uniform/allauth/account/login.html
@@ -1,92 +1,83 @@
 <style>
+    body {
+        background: #f4f6f8;
+        top:0px;
+        font-size: 10pt;
+        margin: 50px 0px;
+        padding: 0px;
+        text-align: center;
+        position: relative;
+        overflow:hidden;
+    }
 
+    button {
+        font-weight: bold;
+        text-transform: uppercase;
+        background-color: #e0e0e0;
+        border: 1px solid #e0e0e0;
+    }
 
+    .inlineLabels {
+        margin-left: -3px;
+        border: 0;
+    }
 
-body {
-	
-    background: #f4f6f8;
-    top:0px;
-    font-size: 10pt;
-    margin: 50px 0px;
-    padding: 0px;
-    text-align: center;
-    position: relative;
-    overflow:hidden;
-}
-button{
-	font-weight: bold;
-	text-transform: uppercase;
-	background-color: #e0e0e0;
-	border: 1px solid #e0e0e0;
-	
-}
-.inlineLabels{
-margin-left: -3px;
-border:0;
-}
+    #div_id_login > label, #div_id_password > label {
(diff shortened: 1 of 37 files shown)


annagav:
Functionality works fine 👍 

One thing I noticed not related to this PR: when I login a student with a course code, I am not directed to the assignments page for that course, but instead I see the SCB homepage and it is not obvious how to get to the assignments.

Another minor (probably not important): when I preview an assignment I created, the 'In the Lab' videos look broken. But this appears to happen only in the preview.

annagav (approved):

jmclaus:
@annagav Thanks for the feedback.

Regarding your first statement, I agree it is confusing and opened an issue, tagged with enhacement: https://github.com/starteam/starcellbio_html/issues/788

Regarding your second statement, you are correct, the 'In the lab' videos slot is empty until a technique is chosen. Seems to happen also, in an identical manner, when I log in as a student. I checked, this behavior is not linked to this PR, so I'll go ahead and merge.

PR #787, opened 2018-01-30
Create new assignment based on another

jmclaus:
This is just https://github.com/starteam/starcellbio_html/pull/760 where merge conflicts were solved. Also, the new `text` and `files` `assignment` fields were added. As the missing `drug.duration_unit` field and `strain_treatment` object in the `copy_assignment` function.

### DIFF
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -281,7 +281,13 @@ def create_assignment(request, course_selected):
         a.name = assignment_name
         a.text = assignment_text
         a.files = assignment_files
-        a.access='private'
+        a.last_page_name = 'techniques'
+        a.facs_last_enabled_page = 'facs_sample_prep'
+        a.micro_last_enabled_page = 'micro_sample_prep'
+        a.has_wb = False
+        a.has_fc = False
+        a.has_micro = False
+        a.access = 'private'
         a.assignment_id = assignment_id
         a.save()
         copy_assignment(based_on, a)
--- a/instructor/common.py
+++ b/instructor/common.py
@@ -279,6 +279,8 @@ def create_assignment(request, course_selected):
         a = deepcopy(based_on)
         a.id = None
         a.name = assignment_name
+        a.text = assignment_text
+        a.files = assignment_files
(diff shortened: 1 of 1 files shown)


annagav (changes requested):
I think there is a bug when you create an assignment based on already published assignment. The created assignment should not be also published.

jmclaus:
@annagav Good catch, thanks. It's fixed, instead of using deepcopy, I now just create a new assignment and copy these fields from the old assignment: `has_concentration, has_temperature, has_start_time, has_duration, has_collection_time `. The new assignment, based on an old one, then has proper default values, like `access` being `private` instead of inheriting for example a `published` value as in the issue you noticed.

annagav:
I think it would be easier to keep the `deep_copy` and just set `a.access='private'`, otherwise there are many other fields missing like `has_wb`.

Also, the Experimental techniques didn't get copied for me.

jmclaus:
@annagav I may have misunderstood the scope of this PR. I thought we wanted to copy the values of the assignment up to the step where the users starts defining techniques. In that case having `has_wb`, `has_fc,` `has_micro` set to their default value of `False` makes sense. Let me just revert this commit and follow your advice.

annagav:
I think the whole assignment should get copied, because the techniques are most time consuming to set up.

jmclaus:
@annagav Well noted, thank.

jmclaus:
@annagav I set the access to private. The left bar inherits but all the sub items in `Western Blotting`, `Microscopy`, and `Flow Cytometry` don't ie we will have a default state for `Lysate Type`, `Sample Prep` etc. So a lot more work has to go in if we want the full assignment to be copied. Perhaps we should settle on the partial solution I first proposed -- it works for the top `Experiment Setup`. And do the rest in another PR. It's been a while I finished my contract hours and I cannot invest too much time.

annagav (approved):
👍

jmclaus:
@annagav Thanks. Just tested it, works fine. Merging.

PR #789, opened 2018-03-13
Strip the url parameters for images

annagav:
Fix image urls
#### What's this PR do?

Strips the parameters in the url of the image when it compiles/publishes a new assignment to the backend `Assignment`.

### DIFF
--- a/instructor/compiler.py
+++ b/instructor/compiler.py
@@ -1,3 +1,4 @@
+import re
 from django.db.models import Q
 from django.utils.html import format_html, format_html_join
 from instructor.models import (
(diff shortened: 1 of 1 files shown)


pdpinch:
Normally I'd prefer to you a library instead of using regular expressions, but maybe that's overkill and I'd like to get this out asap. 

👍

instructor/compiler.py
```diff
@@ -362,7 +363,7 @@ def micro_model(a):
 def generate_slides(assignment):
     ret = {}
     for image in assignment.image.all():
-        ret[image.id] = image.file.url
+        ret[image.id] = re.sub(r'\?[\S]+', '', image.file.url)
```
blarghmatey, line 366: I would use a `.split('?')[0]` rather than a regex here if the intent is just to remove the query parameters. To be more 'correct' I would use urllib, but that's a bit more cumbersome.

pdpinch:
closed in favor of #790

PR #790, opened 2018-03-13
Added logic to remove query params from image url in student view

blarghmatey:
#### What's this PR do?
The student assignment view was unable to load uploaded microscopy images because of an expiration parameter being appended to the S3 URL. This PR strips all query params from those URLs as they are not needed.

`instructor/compiler.py`


pdpinch (approved):
LGTM, altho might be nice to see tests pass

